PR #76, opened 2015-03-31
Speeding up +=  operation of LpAffineExpression

andressommerhoff:
The previous += operation of LpAffineExpression is extremely slow when you have a large quantity of elements inside an LpAffineExpression. This is because the __add__() function was used (and it makes an expensive copy of itself each time a new element was added!). If the += is in a for loop for adding one element each time, it is exponential worst. 

This update implements an __iadd__() function for LpAffineExpression, that avoid copying LpAffineExpression each time new elements are added. Now we can keep the pulp "magic" without performance issue when constructing a big expression (same performance as lpsum() function). For example the following code was previously prohibitive:

    vars = LpVariable.dicts("var",range(1000),0,None,LpContinuous)
    coeffs = range(1000)
    expr = LpAffineExpression()
    for n in range(10):  #Ten times building an expression of 1000 elements
        print n
        for i in range(1000): # 1000 elements
            expr += coeffs[i] * vars[i]

In my cpython it took 13 sec. Now after adding __iadd__ it takes only 0.4 secs!  

PS: An LpContrains also takes the same advantage as it Inheritance the __iadd__ function from LpAffineExpression

### DIFF
--- a/src/pulp/pulp.py
+++ b/src/pulp/pulp.py
@@ -749,11 +749,17 @@ class LpAffineExpression(_DICT_TYPE):
     def __radd__(self, other):
         return self.copy().addInPlace(other)
 
+    def __iadd__(self, other):
+        return self.addInPlace(other)
+
     def __sub__(self, other):
         return self.copy().subInPlace(other)
 
     def __rsub__(self, other):
         return (-self).addInPlace(other)
+
+    def __isub__(self, other):
+        return self.subInPlace(other)
 
     def __mul__(self, other):
         e = self.emptyCopy()


stumitchell:
Right looks good and I'll add it to the next version but I'll have to look at what coin-or's policy on pull requests is (contribution agreement etc)

stumitchell:
so asked and pull requests need to have a contribution agreement, so to avoid the paper work, I will just add this fix myself

vfazio88:
Hello,

I did not understand your upgrade above.
Following up on your example, can you please provide a "faster" way to add constraint in this problem:

import pulp
    vars = pulp.LpVariable.dicts("var",range(1000),0,None,pulp.LpContinuous)
    coeffs = range(1000)
    expr = pulp.LpAffineExpression()
    import time
    start_time = time.time()
    for n in range(1000):  #Ten times building an expression of 1000 elements
        #print n
        for i in range(1000): # 1000 elements
            expr += coeffs[i] * vars[i]
    print("--- %s seconds ---" % (time.time() - start_time))


Thanks in advance

PR #77, opened 2015-04-01
deleting MPS files with COIN_CMD

stumitchell:
From Kylie Hubert

How awesome! Thanks for the great release.

There is one last patch in my local copy that wasn't captured in the latest release:


kjmph:
Thanks

PR #78, opened 2015-04-15
Update tests.py

stumitchell:
added a print statement
This should fail


stumitchell:
This was a demonstration for the Auckland python user group

Issue #74, opened 2015-03-27
why is there no time limit for CPLEX_CMD?


timnon:
CPLEX offers the option to set the time limit when using class CPLEX_CMD, is there a reason why this is not supported? Maybe because other *_CMD classes do not offer this as well?

Add line (for ten seconds):

cplex_cmds += "set timelimit 10\n"

after line:

cplex_cmds = "read "+tmpLp+"\n"

stumitchell:
Hi I have put a code change in https://github.com/coin-or/pulp-or/tree/feature/cplex_cmd-timelimit

as I don't have cplex can you please check that it works

prob.solve(CPLEX_CMD(timelimit=10)

Stu

timnon:
there is a typo and the conversion of self.timelimit to string is missing, the following line fixes this:

cplex_cmds += "set timelimit " + str(self.timelimit) + "\n"

but then it works,
Tim

Issue #73, opened 2015-03-27
encoding of cplex_cmds in class CPLEX_CMD in solvers.py does not work for Python 3


timnon:
A simple fix is to add the line:

cplex_cmds = cplex_cmds.encode('UTF-8')

just before line:

cplex.communicate(cplex_cmds)

This has been tested with Python 2 and 3

stumitchell:
I have done this with the same branch https://github.com/coin-or/pulp-or/tree/feature/cplex_cmd-timelimit can you please check that it works with cplex?

Stu

timnon:
it works with Python 2.7.5.6 and 3.4.1.0

stumitchell:
closed with https://github.com/coin-or/pulp/pull/81

PR #81, opened 2015-04-23
Patch 1

timnon:
issues #73 and #74


timnon:
it would be great if this could make it into the next release, CPLEX is completely not working for Python 3

PR #89, opened 2015-08-29
Update setup.py to use pyparsing2 on Python 2.6 and later

desktable:
Previously pulp only uses pyparsing2 for Python 3. This seems to cause weird dependency problems when I try to install pulp using [conda](http://conda.pydata.org/docs/index.html conda).

This update allows pulp to use pyparsing2 on Python 2.6 and 2.7.


stumitchell:
fixed in https://github.com/coin-or/pulp/commit/01f1f98df194395a6d7e2c59037a4bc0eb3a36b3

PR #93, opened 2015-10-31
Fixed allcombinations() docstr

japborst:
Fixed a tiny, but essential, error in the docstr of allcombinations(), as this shows up in the documentation as well.


stumitchell:
ta

PR #99, opened 2016-02-03
Remove pyparsing version check from setup.py

wikiped:
According to [pyparsing changelog](http://pyparsing.wikispaces.com/News) from July 19, 2013:

 > With release 2.0.1, I've removed the code that was specific to Python 3.x, so that it can be installed on any Python version 2.6 or later.

This makes pyparsing version check redundant as pulp has requirement for python version to be >= 2.6.


tkralphs:
Checks are failing because it looks like you also need to also remove `install_requires = [pyparsing_ver]` from `setup.py`.

PR #105, opened 2016-04-17
Fix CoinMP slacks and tests

tkorvola:
This fixes slack value reporting in COINMP_DLL so that it passes unit tests. That also required some fixes to the tests.


stumitchell:
Ok thanks

PR #107, opened 2016-05-31
Remove pyparsing version check from setup.py

dims:
Variation of:
https://github.com/coin-or/pulp/pull/100

new versions of pyparsing (>=2.0.1) work with python 2.x and 3.x
http://pyparsing.wikispaces.com/News


dims:
@stumitchell hopefully, this is better than PR #100

dims:
@stumitchell We ran into this trying to test co-install of OpenStack congress which uses PuLP. If this PR looks good, please consider a point release as well. thanks!

PR #112, opened 2016-11-13
Added exception handling for GUROBI_CMD readsol()

erikliland:
During  `pulp.pulpTestAll()`, `self.readsol(tmpSol)` in `class GUROBI_CMD` encounters a `ValueError: too many values to unpack (expected 2)` which causes the test to fail.


stumitchell:
Fixed in https://github.com/coin-or/pulp/pull/138

PR #114, opened 2016-11-15
Added exception handling to counteract CPLEX 12.6.3.0 '_dummy'

erikliland:
Added exception handler for the situations where CPLEX calls `assignConsPi(self,values)` and `assignConsSlack(self,values)` with `values` that contain `_dummy`

### DIFF
--- a/src/pulp/pulp.py
+++ b/src/pulp/pulp.py
@@ -1583,16 +1583,22 @@ class LpProblem(object):
 
     def assignConsPi(self, values):
         for name in values:
-            self.constraints[name].pi = values[name]
+            try:
+                self.constraints[name].pi = values[name]
+            except KeyError:
+                pass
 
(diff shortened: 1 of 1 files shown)


rkersh:
It looks like these continuous-integration tests don't actually test all the solvers.  Would it be possible to get "community editions" (or similar) for the commercial solvers installed so that they can be tested?

erikliland:
I agree with `if name != "_dummy":` would be more consistent with `assignVarsVals` and `assignVarsDj`, but this both these checks are vulnerable to other errors og change of behavior in underlying code. 

So I guess the question is:

- Everything with `try: except:`
- Everything with `if != "..."`

What do you think @stumitchell ?

stumitchell:
Yeah i would be nice if it was consistent, but since it has taken me 2 years to merge I will resolve that at a later time.

PR #116, opened 2016-11-27
Prevent flashing windows on Windows

oprypin:
Fixes the problem mentioned in the mailing list: https://groups.google.com/forum/#!msg/pulp-or-discuss/TKBBGF3vOo4/jl6k6_P4CgAJ

I've been using this patch in my public project for about 2 years, and the users never complained.


stumitchell:
Ok makes sense as well

Thanks stu

PR #117, opened 2016-11-27
Don't write duplicate dummy constraints

### LINKED ISSUE #115: Fails on latest GLPK
There are several bugs that I'm running into that previously weren't there. [The code](https://github.com/BlaXpirit/sixcells/blob/81577ee6a925566ce73f567668ef9c392fcf3ecb/solver.py) worked in the past (PuLP 1.20, GLPK 4.55), and still works with the default solver, but not GLPK 4.60.

---
<del>So PuLP runs `glpsol --cuts` and it chokes on this line: `_C139:0 <= 0`, saying:</del>
> <del>...\Local\Temp\1052-pulp.lp:73: missing variable name</del>

<del>I suspect that this originates from `lpSum(list comprehension that happens to be empty) <= 0` and could be avoided in PuLP by discarding such a condition.</del>

---
Another problem with a line `_dummy: __dummy = 0`
> ...\Local\Temp\5876-pulp.lp:73: constraint '_dummy' multiply defined

Probably related to https://github.com/coin-or/pulp/commit/ee88a2964da94a72a754d7e9fe2c564f0eb9a203. I was able to fix this by writing the dummy var only once (setting a boolean flag after the first write).

---
Disclaimer: I don't know what's really going on with the optimization thing here, just maintaining code not written by me.

oprypin:
Fixes #115

### DIFF
--- a/src/pulp/pulp.py
+++ b/src/pulp/pulp.py
@@ -1505,6 +1505,7 @@ class LpProblem(object):
         f.write("Subject To\n")
         ks = list(self.constraints.keys())
         ks.sort()
+        dummyWritten = False
         for k in ks:
             constraint = self.constraints[k]
             if not list(constraint.keys()):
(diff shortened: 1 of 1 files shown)


stumitchell:
ok makes sense

PR #119, opened 2016-12-08
Use thread-safe subprocess32 on Py2

blechta:
https://github.com/google/python-subprocess32 is a backport of Python 3 `subprocess` which is thread-safe on POSIX.

[FFC](https://bitbucket.org/fenics-project/ffc/) (which depends on [COFFEE](https://github.com/coneoproject/COFFEE) which in turn depends on PuLP) is often being run on Infiniband clusters where Python2 `subprocess` implementation is known to randomly segfault. This is the problem related to thread-safety.

### DIFF
--- a/src/pulp/solvers.py
+++ b/src/pulp/solvers.py
@@ -31,7 +31,6 @@ the current version
 """
 
 import os
-import subprocess
 import sys
 from time import clock
 try:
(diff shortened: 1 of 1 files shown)


stumitchell:
Hmm interesting, 
I'll have to have a think about this, probably prefer to wait untill the backport is official, 

can you not use monkey patching to do this yourself??

blechta:
Well, official documentation recommends its usage, see https://docs.python.org/2/library/subprocess.html. It is an official, well-maintained backport. We would prefer this, now standard, solution rather than a monkey-patching.

stumitchell:
Ok you convinced me 😁

stumitchell:
Ok you convinced me 😁

PR #126, opened 2017-03-03
Use tempfile.NamedTemporaryFile to avoid collisions in GLPK_CMD

benmoran:
Because pulp.solvers.GLPKCMD creates temporary filenames using the PID of the process where PuLP is running, then problems created in separate threads in the same process will have problem and solution filename collisions, resulting in errors and incorrect results.

If we use python's tempfile.NamedTemporaryFile instead, we should avoid filename collisions.

### DIFF
--- a/src/pulp/solvers.py
+++ b/src/pulp/solvers.py
@@ -40,7 +40,7 @@ except ImportError:
 from . import sparse
 import collections
 import warnings
-from tempfile import mktemp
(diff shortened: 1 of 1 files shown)


stumitchell:
Thanks for this but this approach seems to fail for coin_cmd so I changed to uuids instead.

There are several tests in the command line solvers that check for the existence of the solution file and NamedTempFile actually creates the file all the time.

Issue #129, opened 2017-04-12
Pulp and Cplex


samvy:
Hi, I try to use pulp with cplex but I have a problem when the package is imported :

Traceback (most recent call last):
  File "test.py", line 1, in <module>
    from pulp import *
  File "/usr/lib/python2.7/dist-packages/pulp/__init__.py", line 34, in <module>
    from .pulp import *
  File "/usr/lib/python2.7/dist-packages/pulp/pulp.py", line 102, in <module>
    from .solvers import *
  File "/usr/lib/python2.7/dist-packages/pulp/solvers.py", line 1001, in <module>
    class CPLEX_PY(LpSolver):
  File "/usr/lib/python2.7/dist-packages/pulp/solvers.py", line 1010, in CPLEX_PY
    CplexLpStatus = {cplex.Cplex.solution.status.MIP_optimal: LpStatusOptimal,
AttributeError: type object 'Cplex' has no attribute 'solution'

Cplex Version = 12.7.1.0
PuLP Version: 1.6.5

Regards

timeregained:
I have met the same issue and the following might be a work around.

From the last minor update from 12.7.0.0 to 12.7.1.0, IBM has delete some contents of __init__.py in cplex package folder.

"
variables             = _internal._subinterfaces.VariablesInterface()
    """See `_internal._subinterfaces.VariablesInterface()` """
    linear_constraints    = _internal._subinterfaces.LinearConstraintInterface()
    """See `_internal._subinterfaces.LinearConstraintInterface()` """
    quadratic_constraints = _internal._subinterfaces.QuadraticConstraintInterface()
    """See `_internal._subinterfaces.QuadraticConstraintInterface()` """
    indicator_constraints = _internal._subinterfaces.IndicatorConstraintInterface()
    """See `_internal._subinterfaces.IndicatorConstraintInterface()` """
    SOS                   = _internal._subinterfaces.SOSInterface()
    """See `_internal._subinterfaces.SOSInterface()` """
    objective             = _internal._subinterfaces.ObjectiveInterface()
    """See `_internal._subinterfaces.ObjectiveInterface()` """
    MIP_starts            = _internal._subinterfaces.MIPStartsInterface()
    """See `_internal._subinterfaces.MIPStartsInterface()` """
    solution              = _internal._subinterfaces.SolutionInterface()
    """See `_internal._subinterfaces.SolutionInterface()` """
    presolve              = _internal._subinterfaces.PresolveInterface()
[...]

PR #134, opened 2017-05-26
Fix issue #129

rkersh:
With CPLEX 12.7.1 the class variables that were used previously for
CplexLpStatus are only available as instance variables. Here, we move
the definition of CplexLpStatus into findSolutionValues. This should
be backwards compatible with older version of CPLEX as well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@ language: python
 python:
 - '2.6'
 - '2.7'
-- '3.2'
 - '3.3'
 - '3.4'
 - pypy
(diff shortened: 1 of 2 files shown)


stumitchell:
Due to a change in setuptools python 3.2 is no longer supported if you could remove this line https://github.com/coin-or/pulp/blob/master/.travis.yml#L5 your tests should pass

rkersh:
I've modified the .travis.yml file as suggested. It looks like the tests are passing now.

stumitchell:
Thanks I'll be doing a new release soon.

Issue #133, opened 2017-05-10
Issue while using pulp with gevent


amit5148:
I am using pulp within a process using gevent.
Intermittently, I am receiving the error from [line 1425 in solvers.py](https://github.com/coin-or/pulp/blob/master/src/pulp/solvers.py#L1425.)
I am assuming that it is caused due to multiple greenlets running concurrently with the same value of tmpSol variable and the greenlet removes the solution file after reading it and causing the error for other greenlets.
I think generating a unique value for tmpSol for individual greenlets should fix the situation.

Am I going the wrong way?
If not, what change would you recommend to fix it?

stumitchell:
let me know if this isn't fixed

PR #137, opened 2017-06-12
Fix issue #133: avoid file naming collisions by using uuids instead of pids

s-c-e:
Hi Stu - I'm a big PuLP fan and figured I could help with this file naming problem (#133 - and I've had the same problem come up in the past). Would you be okay to switch from pids to uuids?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@ language: python
 python:
 - '2.6'
 - '2.7'
-- '3.2'
 - '3.3'
 - '3.4'
 - pypy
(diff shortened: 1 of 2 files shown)


s-c-e:
Per your notes on another PR, I've removed Python 3.2 from the Travis CI file.

s-c-e:
@stumitchell do you think this will fix the issue, or should I try something else?

stumitchell:
No it is good sorry to take so long to put in it. I'll be doing a new release soon

HySoaKa:
Thank you for your pr !!

PR #167, opened 2018-03-21
Feature: Add timeout option for CBC solver

bertrand-caron:
Hi There,

First of all thanks a lot for Pulp, this is very useful and powerful piece of software.

Full disclaimer: This is my first (real) OOS pull-request, so please point out anything I might be doing wrong.

# What this pull request does

Add a `timeout` option for the CBC solver, which relies on `Popen.wait(timeout)`.

# Why

While running `pulp` on a production server (Apache), I found several instances of leftover `cbc` threads which have been running for days. I suspect Apache killed the main Python `CGI` thread but somehow left the `cbc` thread behind. Generally speaking, for problems of unknown size, I think some user might find it convenient to be able to use a timeout as a measure of (immediate) feasibility of a problem.

# Describe the approach

I piggybacked on `Popen.wait(timeout)` which takes an optional `timeout` argument since `3.3` and raises a `TimeoutExpired` exception if exceeding the provided timeout. I simply forwarded the `timeout` keyword argument from high level function (`solve()`, `actualSolve()` and `sequentialSolve()`, for functions not already using keyword argument unpacking) down to `Popen.wait()`. I made sure to make the code `2.7` compatible (obviously they won't be able to use this feature, and will get a warning if they do). I also added a test. I chose not to catch the `TimeoutExpired` and let the user deal with it, rather than wrap it in some specific Pulp exception, but that behaviour can be changed very easily.

I hope you merge it,

Best, Bertrand

### DIFF
--- a/src/pulp/pulp.py
+++ b/src/pulp/pulp.py
@@ -1668,7 +1668,7 @@ class LpProblem(object):
         return status
 
     def sequentialSolve(self, objectives, absoluteTols = None,
-                        relativeTols = None, solver = None, debug = False, timeout = None):
+                        relativeTols = None, solver = None, debug = False, processTimeout = None):
         """
         Solve the given Lp problem with several objective functions.
 
@@ -1680,7 +1680,7 @@ class LpProblem(object):
            the constraints should be +ve for a minimise objective
         :param relativeTols: the list of relative tolerances applied to the constraints
         :param solver: the specific solver to be used, defaults to the default solver.
-        :param timeout: Timeout (in second) passed to Popen.wait() (for CBC solver only)
+        :param processTimeout: Timeout (in second) passed to Popen.wait() (for CBC solver only)
 
         """
         #TODO Add a penalty variable to make problems elastic
@@ -1698,7 +1698,7 @@ class LpProblem(object):
         for i,(obj,absol,rel) in enumerate(zip(objectives,
                                                absoluteTols, relativeTols)):
             self.setObjective(obj)
-            status = solver.actualSolve(self, timeout=timeout)
+            status = solver.actualSolve(self, processTimeout=processTimeout)
             statuses.append(status)
             if debug: self.writeLP("%sSequence.lp"%i)
             if self.sense == LpMinimize:
--- a/src/pulp/pulp.py
+++ b/src/pulp/pulp.py
@@ -1668,7 +1668,7 @@ class LpProblem(object):
         return status
 
     def sequentialSolve(self, objectives, absoluteTols = None,
-                        relativeTols = None, solver = None, debug = False):
+                        relativeTols = None, solver = None, debug = False, timeout = None):
         """
         Solve the given Lp problem with several objective functions.
 
@@ -1680,6 +1680,7 @@ class LpProblem(object):
            the constraints should be +ve for a minimise objective
         :param relativeTols: the list of relative tolerances applied to the constraints
         :param solver: the specific solver to be used, defaults to the default solver.
+        :param timeout: Timeout (in second) passed to Popen.wait() (for CBC solver only)
 
         """
         #TODO Add a penalty variable to make problems elastic
@@ -1697,7 +1698,7 @@ class LpProblem(object):
         for i,(obj,absol,rel) in enumerate(zip(objectives,
                                                absoluteTols, relativeTols)):
             self.setObjective(obj)
-            status = solver.actualSolve(self)
+            status = solver.actualSolve(self, timeout=timeout)
             statuses.append(status)
             if debug: self.writeLP("%sSequence.lp"%i)
             if self.sense == LpMinimize:
--- a/src/pulp/solvers.py
+++ b/src/pulp/solvers.py
@@ -1365,7 +1365,7 @@ class COIN_CMD(LpSolver_CMD):
         """True if the solver is available"""
         return self.executable(self.path)
 
-    def solve_CBC(self, lp, use_mps=True):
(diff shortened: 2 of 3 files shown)


src/pulp/pulp.py
```diff
@@ -1668,7 +1668,7 @@ def solve(self, solver = None, **kwargs):
         return status
 
     def sequentialSolve(self, objectives, absoluteTols = None,
-                        relativeTols = None, solver = None, debug = False):
+                        relativeTols = None, solver = None, debug = False, timeout = None):
```
stumitchell: I would be happer is we used **kwargs to pass solver parameters here

stumitchell:
I would be happy if this was called something more like `processTimeout` as there are timeout options that get passed to cbc.

bertrand-caron:
@stumitchell I've renamed `timeout <- processTimeout`, are you happy with that? Regarding your second comment, should I make it a separate pull request?

amarvin:
Couldn't the existing CBC parameter maxSeconds be used for this?

pchtsp:
I think this could also be extended to other solvers. I'm not sure if it should be the user that gives both the `timelimit` parameter and the `processTimeout` or should the second be calculated from the former plus a tolerance.

@amarvin In my case the `timelimit` (or `maxSeconds` in CBC) parameter is not enough. I don't know why, when I use CPLEX_CMD, if by the time `timelimit` has passed, CPLEX hasn't found any integer solution yet, it decides by its own initiative to restart for the same amount of time (starting where it left the branch and bound exploration). So I get double the amount of solving time for these specific instances. 

What I'm testing is passing the timeout to the `actualSolve` function and using it to kill the process after a reasonable time over the limit has passed.

src/pulp/solvers.py
```diff
@@ -1420,7 +1420,11 @@ def solve_CBC(self, lp, use_mps=True):
         args.append(self.path)
         args.extend(cmds[1:].split())
         cbc = subprocess.Popen(args, stdout = pipe, stderr = pipe)
-        if cbc.wait() != 0:
+
+        if processTimeout is not None and sys.version_info < (3, 3):
+            warnings.warn('Pulp: Subprocess timeout is not available in python < 3.3. Solver will run indefinitely.')
```
rwitzel: The message might be misleading because the solver will not run indefinitely when a timeout is passed to CBC or a solution is found.

Thus I suggest to log `... < 3.3. Passed subprocess timeout is ignored.` instead.

src/pulp/solvers.py
```diff
@@ -1420,7 +1420,11 @@ def solve_CBC(self, lp, use_mps=True):
...
+
+        if (cbc.wait(processTimeout) if sys.version_info >= (3, 3) else cbc.wait()) != 0:
```
rwitzel: As `cbc.wait(timeout)` can raise a `TimeoutExpired` exception, one could consider to catch this exception and raise a `PulpSolverError` instead. But I am not sure which kind of interface is desired here in the project.

src/pulp/pulp.py
rwitzel: @bertrand-caron Sounds reasonable, right?

rwitzel: @stumitchell Using kwargs in all methods, i.e. also in solve_CBC?

rwitzel:
> Couldn't the existing CBC parameter maxSeconds be used for this?

IMHO two different parameters should be used because they might behave differently. Popen's wait might use elapsed time (clock time) and CBC's timeout might limit CPU time, I don't know. The availability of the 'best found solution so far' might differ as well.

hugoantoine:
I am running into a similar issue where I get a problem that when solved as .mps with pulp runs forever, but when written into a .lp format can be solved almost instantly (and in other instances the opposite happens). Any update on this feature ? Because it sounds like it would solve my issue.

src/pulp/solvers.py
hugoantoine: I would suggest this way of catching the _TimeoutExpired_ exception and also killing the cbc subprocess, as it was still able to run after the _cbc.wait(processTimeout)_.
```suggestion
        if processTimeout is not None and sys.version_info >= (3, 3):
            try:
                 cbc.wait(processTimeout)
            except subprocess.TimeoutExpired:
                cbc.kill()
                raise PulpSolverError("Pulp: Error while trying to execute " + \
                                  self.path)
        else:
            if cbc.wait()!=0:
                raise PulpSolverError("Pulp: Error while trying to execute " + \
                                      self.path)
```

lucasklaassen:
We had this exact same use case. I wish I would have seen this PR sooner. Here is the solution we came up with: https://github.com/betterhours/pulp/commit/62bcfa7097a7b4adcc28dbdb28da6115c7acb8cd

Having the ability to set a timeout on these solvers is very useful. It helps save us CPU resources when running production workloads/tests that we don't want running longer than X amount of time.

+1

pchtsp:
Why not just slightly change the cbc solver call so it matches the cplex one and then we can just give the timeout as an argument? It seems a lot simpler and cleaner for me.

This is what I use for the CPLEX solver:

```python
cplex = subprocess.Popen(self.path, stdin = subprocess.PIPE)
# ...
try:
    cplex.communicate(cplex_cmds, timeout=timeout)
except subprocess.TimeoutExpired:
    cplex.kill()

if cplex.returncode != 0:
    # ...
    raise PulpSolverError("PuLP: Error while trying to execute "+self.path)

```

Ideally, we want all interfaces to command line solvers to look as similar as possible.

PR #177, opened 2018-07-04
faster writeMPS

RomanZavodskikh:
use a bit of memory to create strings
        in a single loop before writing to file


pchtsp:
I just tried this approach to writeLP and, with large files. it performed poorly. And when I say poorly I say I didn't even let it finish, after 30 min.

What does work is to create a list of strings and then use:

```
f.write(''.join(constraints))
f.write(''.join(variables))
```
at the end.

example here for `writeLP`: https://github.com/pchtsp/pulp-or/tree/faster_writes

PR #178, opened 2018-07-20
Fix for open pipe

CrystalLord:
PuLP does not close the pipe to /dev/null if it was opened. This causes ResourceWarning:
```python
ResourceWarning: unclosed file <_io.TextIOWrapper name='/dev/null' mode='w' encoding='UTF-8'>
```
This commit closes the pipe if it was opened.


kozako2:
Up

CrystalLord:
A fair number of people have asked for this same fix, since it's so trivial. Unfortunately, it appears the github here is not maintained much.

PR #187, opened 2018-11-30
Change the logic of collecting Gurobi variables to bulk fetching

dinever:
Gurobi 8 started to use HTTP protocal for data transfer, if we call the library to get the value of the variables like this: 

```python
for var in lp.variables(): 
    var.varValue = var.solverVar.X
```
Gurobipy will send out an HTTP request to the gurobi server for every variables. If we have a very large number of variables, `findSolutionValues` can be pretty expensive and slow.

Luckily, Gurobipy also provides a way to bulk fetch all the variables together in a single HTTP request by doing

```python
m.getAttr(GRB.Attr.X, m.getVars())
```

which is a lot faster.

### DIFF
--- a/src/pulp/solvers.py
+++ b/src/pulp/solvers.py
@@ -1827,25 +1827,31 @@ class GUROBI(LpSolver):
                                    GRB.NUMERIC: LpStatusNotSolved,
                                    }
             #populate pulp solution values
-            for var in lp.variables():
-                try:
-                    var.varValue = var.solverVar.X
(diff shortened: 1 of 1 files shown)


spiralulam:
What is the status on that? I've also experienced that PuLP is really slow fetching the results from GUROBI.

dinever:
Hi @spiralulam, looks like pulp is not actively maintained right now. Our team ended up using our own pulp branch.

/cc @stumitchell

PR #193, opened 2019-02-06
Mip start

pchtsp:
Hey Stu!

As promised: this is a tentative PR with MIP start support for CPLEX and CBC (the CMD version both). I have only done a couple of tests with some examples and gotten mixed results (sometimes it recognizes it and use it, sometimes it doesn't). Before continuing I wanted to have your opinion on these changes and the style of the code. As you will see, I've taken the time to refactored some almost-duplicated functions on the way and little modifications like making the `writeLP` function return the variables, in a similar way to the `writeMPS`.

I've realized too late that pulp is supposed to be compatible with python 2.6, where dict comprehensions do not exist. I was about to change it back when I thought Who uses python 2.6 nowadays??

Tell me what you think about it !

Franco


pchtsp:
Hey Stu,

I've:

* implemented the previous "setInitialValue" method in the LpVariable.
* implemented a new "fixValue" method in the LpVariable to fix LpBound and UpBound to the present variable's value.

I've been testing the initial and the fixing methods with my own model and CPLEX and can now say that this is working as expected.

Would you mind if I publish the fork in the pulp google group so as to have people test it? Or do you think I should correct the python 2.6 compatibility errors in order to get the PR into pulp?

I have yet to add the gurobi initial solution functionality. I may do it soon, though.

Franco

stumitchell:
Sure publish the fork of you like. I can't test with cplex.

But yes I do need a 2.6 solution to accept the pr.

Stu

pchtsp:
Ok, fixed the issues. I'm not 100% sure the CHOCO branch is compatible with this one. If you merge any of the two, I can remerge your master with the remaining one to be sure everything is working as expected.

Bouncner:
First of all, thanks for the work! That's something I was desperately looking for! :)

One question: is there a particular reason, that fixValue() does not take the actual value to fix?

pchtsp:
Good day!
I did think of it (in fact I think it was implemented that way in the beginning). So you make a good point.

At the end, for me it was more explicit to do it in two steps. In some user cases, the variables already have a value (from a previous solve, for example) and don't even require the `setInitialValue` function.

Also, I was not sure how to interpret fixing a variable to a value outside its (former) bounds. Should we care if we are going to change the bounds anyways? Should we report it/ ignore it as with `setInitialValue`? I prefered to be sure that if I was fixing a value, I knew it was a value inside the variable domain already.

Having said that, and just for convenience, we could add an optional `val` argument to `fixValue` and just run `setInitialValue` in case it is filled... @stumitchell what do you think?

Bouncner:
I think your reasoning makes sense, but it might be confusing to an outside developer to fix values by first setting an initial value.
Maybe move the validation to another method and both 'fixValue' and 'setInitialValue' call this method when the bounds check is requested?

pchtsp:
I wanted to check how other model languages do.

For `pyomo`, they offer to pass a value as argument to fix. Also, and it's even better, they control the status of the variable (`fixed=True`).

```python
def fix(self, *val):
    """
    Fix the variable. Sets the fixed indicator to
    :const:`True`. An optional value argument will
    update the variable's value before fixing.
    """
    if len(val) == 1:
        self.value = val[0]
    elif len(val) > 1:
        raise TypeError("fix expected at most 1 arguments, "
                        "got %d" % (len(val)))
    self.fixed = True

def unfix(self):
    """Free the variable. Sets the fixed indicator to
    :const:`False`."""
    self.fixed = False
```

I come from using AIMMS for a long time and they had the filling and fixing separated one from the other. Maybe that's why I see it that way.

Lastly, it could be potentially interesting to keep the "real" bounds stored in a backup for when the variable is unfixed. I think AIMMS did this. Not sure about pyomo.

All this to say that the fixing and setting initial values can be greatly improved. This is just a timid first step. Any further changes should be a little more planned.

For now, and as a workaround, you could just creat an auxiliary function like so:

```python
def fix_variable(variable, value):
    variable.setInitialValue(value)
    variable.fixValue()
```

pchtsp:
hey @stumitchell do you think this PR will be merged ? Do you need anything else ?

thanks !

Franco

stumitchell:
@pchtsp can you fix these conflicts

juanpineros:
Hi Franco. I need to do Fix & optimize heuristic and I used your help about setinitialvalue for it will be start with the before solution. But when I run the heuristic the .log of cplex didn't identify the first solution 'm1'.

My line of solve is:
Pulp.solve(CPLEX())

What are you doing wrong ?, I need to use other command?
Thanks

pchtsp:
Hey!

Check this guide here:

https://coin-or.github.io/pulp/guides/how_to_mip_start.html

juanpineros:
Thanks for your information Franco,
I Tried with the information of example but I continue with the problem. Now the problem is:

  File "C:\Users\Admin\Anaconda3\lib\site-packages\pulp\solvers.py", line 1569, in __init__
    COIN_CMD.__init__(self, path=self.pulp_cbc_path, *args, **kwargs)

TypeError: __init__() got an unexpected keyword argument 'mip_start'

Also, I tried with your code "wedding_initial" but I had the same error.

pchtsp:
Hello, I think you need to update your pulp version.

By doing

```pip install - U pulp```

And check it is the version 2.0

F.

juanpineros:
Hello Franco, thnaks for your help I had last  pulp version. Now I use to initialValue and the corrects parameters for mip start, but I want fix some variables with the initial value and use fixValue() but I have the next erro.

My code is>
            for k, v in varsdict.items():
                k.setInitialValue(v)
                k,fixValue()

The erro is>
 File "C:\Users\Admin\Anaconda3\lib\site-packages\pulp\pulp.py", line 1713, in solve
    status = solver.actualSolve(self, **kwargs)
 
  File "C:\Users\Admin\Anaconda3\lib\site-packages\pulp\solvers.py", line 526, in actualSolve
    status, values, reducedCosts, shadowPrices, slacks = self.readsol(tmpSol)

  File "C:\Users\Admin\Anaconda3\lib\site-packages\pulp\solvers.py", line 555, in readsol
    raise PulpSolverError("Unknown status returned by CPLEX: "+statusString)

PulpSolverError: Unknown status returned by CPLEX: time limit exceeded

Thank.

pchtsp:
That's a weird error. Could you open a new issue with this please and
giving as much information as you can ?

Thanks!

juanpineros:
Hi Franco,
I'm use to parameters  initialValue and fixValue but the answer are different and I fixed all the variable.

            for k, v in varsdict.items():
                k.setInitialValue(v) 
                k.fixValue()

solver = pulp.CPLEX_CMD(msg=1, mip_start=1, timelimit = time_limit,keepFiles = True)
prob.solve(solver)

If thinking about all variables are fixed só the answer must be the same for all interaction.

Answer:
run 1: 1,9E+08
run 2: 2,02E+08
run 3: 2,02E+08
run 4: 1,96E+08
run 5: 1,9E+08
run 6: 2,02E+08

Can you help me, what is my mistake ?

pchtsp:
Please, open an issue on this repository or write to the Google group at https://groups.google.com/forum/m/#!forum/pulp-or-discuss . 

This is not the correct channel for support.

Thanks!

PR #195, opened 2019-02-25
Update pulp.py

jakedaford:
Current code currently overwrites constant in lhs with negative of rhs instead of subtracting rhs... seems like a bug

### DIFF
--- a/src/pulp/pulp.py
+++ b/src/pulp/pulp.py
@@ -852,7 +852,7 @@ class LpConstraint(LpAffineExpression):
         """
         LpAffineExpression.__init__(self, e, name = name)
         if rhs is not None:
-            self.constant = - rhs
+            self.constant -= rhs
         self.sense = sense
         self.pi = None
         self.slack = None


jakedaford:
Just to include an example

```
>>> import pulp
>>> x = pulp.LpVariable('x')
>>> expr = pulp.LpAffineExpression({x: 1}, constant=3)
>>> c = pulp.LpConstraint(e=expr, rhs=5)
>>> c
1*x + -5 = 0
```

stumitchell:
Interesting, I never used the constant arguement on lpAffineExpression.
I suspect you are right, Will ponder this for a bit, do you have a usecase or were you just looking at the code.

jakedaford:
In example problems about costs, we could have
cost_run_machine_1 = 100 + 50x ($100 to start it up then $50 an hour)
cost_run_machine_n....
total_cost = sum(costs) <= 5,000

Also, even if the AffineExpression wasn't instantiated with a constant it could pick one up along the way... basically, as written, the user must "collect constant terms" before creating constraints.

stumitchell:
can you add a test for this in https://github.com/coin-or/pulp/blob/master/src/pulp/tests.py

jakedaford:
will do, but probably won't be til the weekend or so

dinever:
Hi @stumitchell, thanks for merging this! Would you mind bumping the version on https://pypi.org/project/PuLP/ once you get a chance? Thanks a lot!

stumitchell:
Should be up now give it a go

Stu

Stuart Mitchell
PhD Engineering Science
Extraordinary Freelance Programmer and Optimisation Guru
www.stuartmitchell.com

dinever:
thanks a lot!

PR #202, opened 2019-06-14
Update solvers.py for solve(GUROBI()) and solve(CPLEX(mip = False))

PanZhanhua:
GUROBI: In python3.7 and gurobi-python37, the shadowPrice(or Pi) in gurobipy.GRB is not GRB.Pi but GRB.Attr.Pi. Same as slack. Other earlier versions have not been tested.

CPLEX: In the class CPLEX_PY, before "put pi and slack variables", "lp.solverModel.get_problem_type" is a function, so it should be "lp.solverModel.get_problem_type()". And we can use "prob.solve(CPLEX(mip = False))" when we know the prob is a LP problem and we want to konw the Pi and slack.

### DIFF
--- a/src/pulp/solvers.py
+++ b/src/pulp/solvers.py
@@ -1191,7 +1191,7 @@ else:
                 lp.assignVarsVals(variablevalues)
                 constraintslackvalues = dict(zip(con_names, lp.solverModel.solution.get_linear_slacks(con_names)))
                 lp.assignConsSlack(constraintslackvalues)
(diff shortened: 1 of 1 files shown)


pchtsp:
Thank you for the PR.
Can you check if there are existing unittests that test shadow price extraction for these two solvers? And if not, add them please? thanks

pchtsp:
Thanks for the PR. I've incorporated the changes and also they are tested now since I took out the try/ excepts in the api. So I'll close the PR.

PR #206, opened 2019-07-14
Use Gurobi's OS X location in the OS X config file

ludwiktrammer:
The previous value pointed to the default location specific to Linux


pchtsp:
I'm just closing old PRs. This PR was too small I guess.

ludwiktrammer:
I don't get it. The bug that the PR is fixing still exists - the path set for Mac OS X is incorrect.

pchtsp:
given that the path to gurobi changes with each version of gurobi, the path is and will always be incorrect. Yours is only valid for version "gurobi811".
Instead of using the pulp config file I recommend you use environment variables. As explained here:
https://coin-or.github.io/pulp/guides/how_to_configure_solvers.html#configuring-the-path-to-the-solver

Issue #192, opened 2019-02-06
Deprecation warnings in version 1.6.9, will stop working with Python 3.8


larsblumberg:
Using the current version 1.6.9., py.test spits out the following warning which indicates that PuLP will stop working with the upcoming Python version 3.8:

```
venv/lib/python3.7/site-packages/pulp/pulp.py:1672: DeprecationWarning: time.clock has been deprecated in Python 3.3 and will be removed from Python 3.8: use time.perf_counter or time.process_time instead
    self.solutionTime += clock()
```

tirkarthi:
`time.clock` is now removed in 3.8 : https://github.com/python/cpython/pull/13270/

PR #211, opened 2019-08-06
Also run unittests for Python versions 3.7 and newer in Travis CI

bjk17:
In order to catch issues like #192 (library deprecated in Python 3.8) as early as possible the unittests should always be run for all supported Python versions in this library's CI loop. Therefore I propose adding Python versions `3.7`, `3.7-dev`, `3.8-dev` and `nightly` (along with PyPy 3.5) to the Travis CI loop.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,26 +1,47 @@
 language: python
-python:
-- '2.6'
-- '2.7'
-- '3.3'
-- '3.4'
-- '3.5'
-- '3.6'
-- '3.7'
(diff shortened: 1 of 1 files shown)


bjk17:
Then I suggest releasing PuLP version `1.6.11` with the fix for Python 3.8 as the newest version `1.6.10` on PyPi is incompatible with Python 3.8.

stumitchell:
Great thanks

PR #214, opened 2019-08-20
The "Infeasible" result by the CPLEX solver is now detected correctly.

progirep:
This was mentioned as a TODO.

### DIFF
--- a/src/pulp/solvers.py
+++ b/src/pulp/solvers.py
@@ -545,6 +545,7 @@ class CPLEX_CMD(LpSolver_CMD):
         # TODO: check status for Integer Feasible
         cplexStatus = {
             "optimal":LpStatusOptimal,
+            "infeasible":LpStatusInfeasible
             }
         if statusString not in cplexStatus:
             raise PulpSolverError("Unknown status returned by CPLEX: "+statusString)


pchtsp:
Hello!

Thanks for the initiative.
Unfortunately, what we wanted is to differentiate between integer feasible
solutions and optimal. Infeasible solutions are already detected by the
lack of solution file.

I’m not sure if CPLEX returns the status for integer feasible but if it
did, we would use it.

For example, in the case of CBC we do the following to detect integer
feasible or optimal solutions (via status and sol_status):

https://github.com/coin-or/pulp/blob/183dbdff3f6a4e559af8d503752c27c95b0c2a00/src/pulp/solvers.py#L1548

Franco

progirep:
I actually ran into the problem that an exception was thrown at line https://github.com/coin-or/pulp/blob/183dbdff3f6a4e559af8d503752c27c95b0c2a00/src/pulp/solvers.py#L550 because CPLEX returned *in the solution file* that there was no solution. The model only utilized real-valued variables. So while you are certainly right that the TODO comment should not be removed, the one-line fix may still make sense to be added.

I'm seeing this behavior with CPLEX 12.8.0.0.

pchtsp:
makes sense. I guess, undo the TODO line and it should be fine.

@stumitchell what do you think?

progirep:
Thanks for your reply. I'm closing this pull request as I've pushed the amended change request to a new pull request: https://github.com/coin-or/pulp/pull/251

PR #217, opened 2019-08-29
Call SCIP from path defined in config file

roessig:
SICP is currently not found although specified in the config file.


pchtsp:
Thanks for the help.
1. Is there any difference between using `'%s' % self.path` and just `self.path` ?
2. Could you please resolve the conflicts so we can merge this?

Thanks again

roessig:
I think it did not work with just `self.path`, but cannot say why anymore. As far as I can see I cannot resolve the conflict, as I don't have the permission.

pchtsp:
You need to merge the pulp master into your branch and correct the errors, then do a push to your fork.

pchtsp:
Or just create a clean new pull request from the current master.

pchtsp:
I'm closing this but please create a new PR with the change.

PR #220, opened 2019-09-24
Replace the deprecated os.popen with subprocesses for XPRESS

viggotw:
Replaced the deprecated os.popen with subprocess.Popen for the Xpress.actualSolve()-method

### DIFF
--- a/src/pulp/solvers.py
+++ b/src/pulp/solvers.py
@@ -1230,9 +1230,11 @@ else:
 
 class XPRESS(LpSolver_CMD):
     """The XPRESS LP solver"""
+    csv = __import__('csv')
+
     def __init__(self, path = None, keepFiles = 0, mip = 1,
             msg = 0, maxSeconds = None, targetGap = None, heurFreq = None,
-            heurStra = None, coverCuts = None, preSolve = None,
+            heurStra = None, coverCuts = None, preSolve = None, fullPrecision = False,
             options = []):
         """
         Initializes the Xpress solver.
@@ -1255,6 +1257,7 @@ class XPRESS(LpSolver_CMD):
         self.heurStra = heurStra
         self.coverCuts = coverCuts
         self.preSolve = preSolve
+        self.fullPrecision = fullPrecision
 
     def defaultPath(self):
         return self.executableExtension("optimizer")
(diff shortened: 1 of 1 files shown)


viggotw:
For more information:
https://groups.google.com/forum/#!topic/pulp-or-discuss/G-1_AGxSWkA

src/pulp/solvers.py
```diff
@@ -1275,75 +1278,94 @@ def actualSolve(self, lp):
...
             except: pass
+        lp.status = status
         lp.assignVarsVals(values)
-        # TODO: the following could be put inside the assignStatus function
         if abs(lp.infeasibilityGap(self.mip)) > 1e-5: # Arbitrary
-            status = LpStatusInfeasible
-        lp.assignStatus(status)
-        return status
+            lp.status = LpStatusInfeasible
+        return lp.status
```
pchtsp: you need to keep lp.assignStatus here instead of directly assigning lp.status a value

src/pulp/solvers.py
```diff
@@ -1275,75 +1278,94 @@ def actualSolve(self, lp):
...
-            while 1:
-                l = f.readline()
-                if l == "": break
-                line = l.split()
-                if len(line) and line[0] == 'C':
-                    name = line[2]
-                    value = float(line[4])
-                    values[name] = value
-        return status, values
+        if self.fullPrecision:
```
pchtsp: I think this is better if you make a new function (`readsol_fullprecision`) and leave `readsol` as it is.

src/pulp/solvers.py
```diff
@@ -1230,9 +1230,11 @@ def actualResolve(self,lp):
 
 class XPRESS(LpSolver_CMD):
     """The XPRESS LP solver"""
+    csv = __import__('csv')
```
pchtsp: I think the import statement should be passed to the top so it is explicit. At some point we will break the files into different files but for now it is good to have every import at the top.

src/pulp/solvers.py
```diff
@@ -1275,75 +1278,94 @@ def actualSolve(self, lp):
...
+            xpress.stdin.write("GLOBAL\n")
+        if self.fullPrecision:
+            xpress.stdin.write("WRITESOL "+tmpSol+ " -p\n")
+        else:
+            xpress.stdin.write("WRITEPRTSOL "+tmpSol+"\n")
+        xpress.stdin.write("QUIT\n")
+        xpress.stdin.close()
+        if xpress.wait() != 0:
             raise PulpSolverError("PuLP: Error while executing "+self.path)
         status, values = self.readsol(tmpSol)
```
pchtsp: in fact, you can do the `if` statement here and use the two functions directly (`readsol` and `readsol_fullprecision`) depending on the value of `fullprecision`. This way you do not need to assign it to the object to pass it through.

PR #225, opened 2019-10-06
(incomplete) re-organized directories and testing scripts.

pchtsp:
I've taken out the src directory and moved all pulp tests to the `test.py` script (before they were splitted between `tests.py` and `pulp.py`. And moved in under the `tests/` folder in the root (where `amply_tests.py`) was already. I still need to change the pointers to the new location of the scripts for travis to find them. How do you do that?. What do you think about these changes?

I've also edited `ampy_tests.py` to use the `unittests` library. Maybe one day I'll do that with the pulp tests too.


willu47:
#229 might be an easy addition for you?

pchtsp:
In this PR Amply tests are included in automatic tests already, so yes it
should be easy : )

We test pulp, amply and we run examples. Take a look at file
tests/run_tests.py for more details.

PR #230, opened 2019-10-16
Fix for failing Amply tests - Issue 229

willu47:
A small fix which corrects the failing Amply tests.

The `default` argument was returned as a tuple rather than a single value.


src/pulp/amply.py
```diff
@@ -691,7 +691,7 @@ def mark_transposed(tokens):
     param_record = param_slice_record | plain_data_record | tabular_record | \
             tr_tabular_record | Suppress(":=")
 
-    param_default = Optional("default" + data.setResultsName('default'))
```
pchtsp: in #225 we had made the change from `data` to `number` instead of `single` in order to pass the tests and it worked. I'm no expert on pyparsing. Do you know which one of thw two is more appropiate in the context of this code?

PR #234, opened 2019-10-31
added action to deploy to pypi when tagged commit is sent.

pchtsp:


.github/workflows/publish-to-test-pypi.yml
```diff
@@ -0,0 +1,27 @@
...
+    name: Build and publish Python 🐍 distributions 📦 to PyPI and TestPyPI
+    runs-on: ubuntu-18.04
+    steps:
+    - uses: actions/checkout@master
+    - name: Set up Python 3.7
+      uses: actions/setup-python@v1
+      with:
+        python-version: 3.7
+    - name: Build a binary wheel and a source tarball
+      run: python setup.py sdist bdist_wheel
```
stumitchell: I always did `sdist` only and did it with a zip format

pchtsp: so just `python setup.py sdist` ?

PR #235, opened 2019-10-31
Take out highlighting in README.rst and correct an error caused by ad…

pchtsp:
…ding the license text as an argument in setup.py (https://github.com/pypa/twine/issues/454)


pchtsp:
@stumitchell  Ok, apparently, it deploys correctly this time (tested in my repo). Do you want me to change the `python setup.py sdist bdist_wheel` into `python setup.py sdist` ?

stumitchell:
`python setup.py sdist --formats=gztar,zip` 
should be fine I merged this with master as I couldn't see the actions

PR #237, opened 2019-10-31
Change clock() to process_time() for python 3.8 support. fixes #232

### LINKED ISSUE #232: clock() is deprecated
In pulp/pulp.py on lines 1680, 1682, 1712, and 1724 a call to clock() is made. 
time.clock has been deprecated in Python 3.3 and will be removed from Python 3.8

Jacob-harris-94:


Jacob-harris-94:
To whoever looks at this, I assume a conditional check on the python version isn't the best way to solve it. I don't know a better way, do you have any suggestions?

pchtsp:
Just checked the code. In fact, the clock function is indirectly imported from `solvers.py` via the `from .solvers import *` statement (which I think is not great and maybe something to change).

Either way, in `solvers.py` there is the following import:

    try:
        from time import clock
    except ImportError:
        from time import perf_counter as clock

So there's already some compatibility being taken into account.
Is that supposed to fix the issue? I'm not sure what the difference between `perf_counter()` and `process_time()` is. But, in any case, one option could be to just to slightly change that part into:

    try:
        from time import process_time as clock
    except ImportError:
        from time import clock
        
or:

    try:
        from time import process_time as clock
    except ImportError:
        from time import perf_counter as clock

Whatever should be the alternative. What do you think?

Jacob-harris-94:
Interesting... `perf_counter` was **not** deprecated, so I would have expected it to work just fine, now that I know the import for `clock` is actually coming from `solvers.py`. Perhaps that same import should be brought into `pulp.py`:

```python
try:
    from time import process_time as clock
except ImportError:
    from time import clock
```

pchtsp:
I'm not sure how I feel about duplicating a compatibility import, specially if it's not the same code. I think we need to decide whether to use process_time or perf_counter first, and then in which file to put this try catch. Afterwards, we need to explicitly import it from the other file. Like "from X import Y" instead of using the wildcard.

Jacob-harris-94:
`perf_counter` and `process_time` are both fractional seconds.
`perf_counter` is highest available resolution, and "...include time elapsed during sleep and is system-wide."
`process_time` "... does not include time elapsed during sleep". 
**Both** say "The reference point of the returned value is undefined, so that only the difference between the results of consecutive calls is valid.".

I didn't see any instances of sleep used, so I'm not sure it would matter. 
Agreed, regarding only putting it in once place.

https://docs.python.org/3/library/time.html

pchtsp:
the changes were not correctly done and comments were not addresed. In fact, we have already change the imports so they do not give a warning.

Issue #155, opened 2017-10-27
Allowing Multi-threaded LpVariable construction


JulienBerducat:
Hi,

I am currently working on a business problem that requires a thorough graph exploration to build relevant LpVariable for pulp problem.
As the size of the problem grew, this step was taking longer and longer and we decided to go parallel as we have naturally independant variables sets (typically a dozen or so, hence as many as CPU cores on our computers). Parallelism was implemented using multiprocessing.Pool under Python 2.7.

Unfortunately, I experienced some random crash **after** having runned pulp: Graph exploration was running fine (and multiprocessing exited properly) but I faced stackoverflows here and then when building my MIP problem later on: 
* sometimes when building constraint A
* sometimes when building constraint B 
* or even when defining problem's objective function...

Eventually I found out that my issue stem from the **hash function** of LpElement: it is defined from the id() function that returns the memory address holding the LPVariable. 
When runned in parallel, it might happened that different variables will hold the same hash value. Later on, if these variables were added to the same constraint or objective function, LPAffineExpression's internal dict had to rely on LpElement::__eq__ operator to store them properly, which in turn caused the StackOverflow on LpAffineExpression having a few hundreds of items.

I propose a pull request that solves that issue for me, but I guess familiar pulp developpers may consider other way of fixing it. As long as this can go in the master branch someday and help other people run pulp variable creation in paralle, I will be glad :)

Cheers,

Julien

JulienBerducat:
Any chance this could be included in a future release? :)

pchtsp:
if you make a PR we will check it out. (sorry for the late reponse)

JulienBerducat:
Hello Franco,

Better late than never! Many thanks for your reply :) 
Here is the PR: #247 

Sorry if it looks a bit messy, I am not yet very used with contributing to github.

Cheers,

Julien

amakarowsky:
Hi all,

I am also experiencing stack overflow on LpAffineExpression. Multithreading leading to hash collisions in the underlying constraint setup appears very plausible as the cause: I have tried all manner of variations in list comprehensions for constraints, separation by different rows etc.

If there are issues with Julien's PR, are there any other efficient means of tackling this problem? It is producing substantial instability for my application using PuLP.

anabapy:
now,Multi-threaded  solve can we use it?

anabapy:
now,Multi-threaded  solve can we use it?

anabapy:
now,Multi-threaded  solve can we use it?

[...]

PR #247, opened 2019-11-21
Allowing Multi-threaded LpVariable construction #155

JulienBerducat:
Avoid calling LpAffineExpression constructor from LpElement::eq() to avoid stack overflow on comparison operators in case of hash collision of LpElement items.

Note that I considered using a more advanced hash function for LpElement (using mmh3 for instance) possibly based on name attribute... but this is another improvement that I am not sure is relevant here.

The main advantage of this fix is that it allowed me to use pulp on several thread during LpVariable construction and avoid the crashed stemming from variables with same hash functions once all thread were merged.

### DIFF
--- a/pulp/pulp.py
+++ b/pulp/pulp.py
@@ -230,7 +230,10 @@ class LpElement(object):
         return LpAffineExpression(self) >= other
 
     def __eq__(self, other):
-        return LpAffineExpression(self) == other
+        if isinstance(other, LpElement) and self.hash == other.hash:
+            return not self.__ne__(other)
+        else:
+            return LpAffineExpression(self) == other
 
     def __ne__(self, other):
         if isinstance(other, LpVariable):


pulp/pulp.py
```diff
@@ -230,7 +230,10 @@ def __ge__(self, other):
         return LpAffineExpression(self) >= other
 
     def __eq__(self, other):
-        return LpAffineExpression(self) == other
+        if isinstance(other, LpElement) and self.hash == other.hash:
+            return not self.__ne__(other)
```
pchtsp: I'm not sure why you're doing this but this i think this does not do what you want it to do.

`__ne__` returns a boolean value (becuse "!=" constraints do not exist in MIP problems).
`__eq__` returns an `LpAffineExpression.__eq__` function, which itself returns a `LpConstraint`.

So... you cannot return a boolean when you should be returning a `LpConstraint` object.

JulienBerducat: Indeed... I clearly missed that point.
However  __eq__ function will still be called in case you are storing severel LpVariable in a dict that have some hash collision... causing the StackOverflow based on current __eq__ implementation returning a new object...
Perharps a new definition of the hash function would be a better fix?

stumitchell:
Do you really need to generate variables in different threads, in my experience model building should be much faster than model solving.
Perhaps you should look at other ways to optimise you code

JulienBerducat:
Hello Stuart,

If you talk about a classical approach (full model -> solve -> results), I
agree that the solving part is more likely to be the bottleneck.
But if you plan to generate variable candidates in a smart preprocessing,
then this task is likely to be time consuming.
From my experience, this is very helpful for large scale problems as it
allows you to separate the feasibility problem (finding candidates
variables) and the optimality one (which is likely to be simplified a lot
once stripped from feasibility constraints).

Of course you can always build your own data structure during the
preprocessing in multithreading and then consolidate them later one while
modelling... But i don't see any logical reason to prevent creating
LpVariable from the start. Especially if the root cause is the recourse to
the id() function for LpElement::__hash__()



Le dim. 1 déc. 2019 à 20:57, Stuart Mitchell <notifications@github.com> a
écrit :

pchtsp:
I agree with you on the second paragraph. Although I'm not sure if the
changes can be done, I usually prefer to do preprocessing outside PuLP and
then build the model only just before solving.

That way you have a lot more flexibility on the structure of the data that
you will use to do the preprocessing, and probably also can take advantage
of more efficient libraries for data manipulation (pandas, numpy, scikit,
graph-tool, etc.).

F.

amakarowsky:
Variable generation with across different threads appears to be occur during list comprehensions in recent versions of Python (3.7.4 in my experience). Whether we proceed with this patch or another, the issue is proving highly disruptive for normal users who are not actively using thread() etc.

I am not sure if the solution proposed here is perfect, but it did work for me in resolving stack overflow with LpAffineExpression. I would like to thank to @JulienBerducat for sharing it.

pchtsp:
Hello, thanks for the feedback.

Could you please share some reproducible example we can check?

This issue should not appear unless you are trying to make parallel
creation of constraints, which makes very little sense for most
applications.

So the fact that you are having this error may mean that you are doing
something you are not supposed to do.

Thanks!

Franco

amakarowsky:
Hi Franco,

Thank you for getting back to me, I really appreciate it.

Unfortunately I cannot share my exact LP or program. I can however highlight this thread of stack overflow which may be related to our issue here: https://stackoverflow.com/questions/50030498/recursion-error-using-pulp 

To be clear: I am not deliberately attempting to create constraints in parallel. Are we completely sure there are no other scenarios related to the equality check in LpAffineExpression that could lead to a hash collision?

Once again, I appreciate you looking into this. AM

pchtsp:
All I can say for that thread is that I 100% agree with the response. The
way it creates objective functions and constraints is very non-standard.
I'd even say wrong. And it is not reproducible either way.

Sorry, but without any example (even a small, invented one) we cannot look
it further. I cannot guarantee about the impossibility but I've never had
that problem before.

F.

stumitchell:
Unfortunately, I have to agree with Franco, asking us to be 100% sure there are no problems is difficult, but we can guarantee  that if you spend some time finding a small instance reproducible instance of the problem we will solve it.

> Variable generation with across different threads appears to be occur during list comprehensions in recent versions of Python (3.7.4 in my experience).

I don't believe this is true can you provide a reference to indicate that this is happening?

https://stackoverflow.com/questions/50030498/recursion-error-using-pulp

This error doesn't have anything to do with parallel execution and in fact is a recursion error with non-typical use of pulp. Unfortunately I cannot replicate it with the current code.

@amakarowsky please come up with a small example.

stumitchell:
@JulienBerducat There are some problems with using with the hash function and LpVariables because internally we store constraints as dictionaries keyed by the LpVariable with the coefficient as the value.

Ideally we would not allow a LpVariable to be a dictionary key as it is a mutable type like a List i.e. {[]:[]} gives an error

```
>>> {[]:[]}
TypeError: unhashable type: 'list'

>>> {():{}}
{(): {}}
>>> 
```

So we wont be able to include your patch at the moment but perhaps we can refactor pulp not to need the hash function on lp variables.

PR #251, opened 2019-11-26
Infeasibility is now correctly detected for CPLEX (for problem instan…

progirep:
This is the fixed version of an [earlier pull request](https://github.com/coin-or/pulp/pull/214) to make the detection of the "infeasibility" result possible for the CPLEX backend. The old pull request was subsequently closed.

### DIFF
--- a/pulp/solvers.py
+++ b/pulp/solvers.py
@@ -550,6 +550,7 @@ class CPLEX_CMD(LpSolver_CMD):
         # TODO: check status for Integer Feasible
         cplexStatus = {
             "optimal":LpStatusOptimal,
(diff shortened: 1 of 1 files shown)


pchtsp:
Hello! Thanks for contributing.

Can you please provide a test case and add it to the test set that didn't
work before and is fixed with this change?

Thanks,

Franco

progirep:
Was this fixed? Sorry for not being able to provide a test case - this proved to be quite complicated as the non-working case is deeply embedded in another application.

pchtsp:
Since there was no test case, I cannot be sure this change does anything
useful...

PR #252, opened 2019-12-02
Example problem using scenarios. 

lluang:
I created a stochastic optimization problem for the library.  The problem example taken from Data, Models, and Decisions by Bertsimas and Freund, 2nd Edition ISBN: 097591460X


pchtsp:
Hello!

Thanks, that's great. Do you feel like adding a longer description and in the format of [CaseStudies](https://github.com/coin-or/pulp/tree/master/doc/source/CaseStudies) and add it to the docs ?


Franco

lluang:
Ok, I added an .rst example into the Case studies folder. Does that look like what you are looking for. Note: I don't know how to test the include from files lines, but I think I understand them.

PR #256, opened 2020-01-08
Change to subprocess.Popen in CBC solver

ggsdc:
When using python, as a subprocess,  from R, in Windows, with the library reticulate if we use the CBC solver we get the following error:

> OSError: [WinError 6] The handle is invalid
>    File "C:\Program Files (x86)\Python35-32\lib\subprocess.py", line 911, in __init__
>    File "C:\Program Files (x86)\Python35-32\lib\subprocess.py", line 1117, in _get_handles

The solution was found in: [here](https://stackoverflow.com/questions/40108816/python-running-as-windows-service-oserror-winerror-6-the-handle-is-invalid)

Just have to set `stdin = subprocess.DEVNULL` in the call to cbc. Have been using with these change for a while and haven't found any error.


ggsdc:
The original problem was in windows, both 32 and 64 bits, and different OS.

There is already an if statement with the python version in the import for the subprocess library, loading different versions of it depending on the python version, I could create a constant there that is used by cbc later on.

I have not seen any other impact and could not find the same error with Gurobi (the any other solver I could have access to).

Karthik1904:
Hi,
I'm using windows server, getting below error
OSError: [WinError 6] The handle is invalid

I checked below in coin_api.py, it was correct.
stdin = subprocess.DEVNULL

Please suggest

AhmetZamanis:
Hi,
I am using Python through R with the reticulate package and I still have the same error.
Below is the traceback for `m.solve()`

```
  File "<string>", line 2, in <module>
  File "C:\Users\PC\DOCUME~1\Work\DATASC~1\GitHub\USEDCA~1\venv\lib\site-packages\pulp\pulp.py", line 1913, in solve
    status = solver.actualSolve(self, **kwargs)
  File "C:\Users\PC\DOCUME~1\Work\DATASC~1\GitHub\USEDCA~1\venv\lib\site-packages\pulp\apis\coin_api.py", line 137, in actualSolve
    return self.solve_CBC(lp, **kwargs)
  File "C:\Users\PC\DOCUME~1\Work\DATASC~1\GitHub\USEDCA~1\venv\lib\site-packages\pulp\apis\coin_api.py", line 202, in solve_CBC
    cbc = subprocess.Popen(args, stdout=pipe, stderr=pipe, stdin=devnull)
  File "C:\Program Files\Python310\lib\subprocess.py", line 837, in __init__
    errread, errwrite) = self._get_handles(stdin, stdout, stderr)
  File "C:\Program Files\Python310\lib\subprocess.py", line 1290, in _get_handles
    c2pwrite = _winapi.GetStdHandle(_winapi.STD_OUTPUT_HANDLE)
OSError: [WinError 6] The handle is invalid
```
I tried modifying line 202 to have `stdin=subprocess.DEVNULL ` (by default it was `stdin=devnull` as you can see), but this still results in the same error and traceback.

I can try using another solver for now, but let me know if you need more information from me regarding this issue. Thank you.

Issue #260, opened 2020-01-31
Inconsistent solver arguments across solver types, quality of life suggestion


seangrogan:
Hello,

So I am running my code across computers that may or may not have certain solvers installed.  I created a function to get "best available solver", i.e. this:

    def _get_best_solver():
        solvers = [
        pulp.GUROBI,
        pulp.GUROBI_CMD,
        #etc 
        # solvers in the order of preference
        ]
        for solver in solvers:
            if solver().available():
                print(f"Solver {solver} is available")
                return solver
        assert False, "No solver found"`

    def solve_mip():
        #build model
        solver = _get_best_solver()
        model.solve(solver(msg=True, timelimit=300))

I want to pass a time limit of (say) 300 seconds.  The problem is that the solver class initializers seem to have different methods of assigning a time limit.  E.g. `CPLEX_CMD` has `timelimit`, `GUROBI,` `CPLEX_DLL`,  and `CPLEX_PY` has `timeLimit` (capital L), `XPRESS`, `COIN` has `maxSeconds`.  

I'm not familiar with how to do pull requests, but I'd be willing to help make the changes

stumitchell:
Yes this is annoying, I would love a pull request,

Please make the old arguments still work however maybe with a deprecation
warning.

Stu
Stuart Mitchell
PhD Engineering Science
Extraordinary Freelance Programmer and Optimisation Guru
www.stuartmitchell.com

pchtsp:
Maybe it is too ambitious but if you want to make a proposal of a solver
factory, that would be great too.

I attach an example of doing this (making arguments more standard and also
calling a solver with a string) that I use in some projects of mine.

tkralphs:
Attachments get scrubbed when you respond by e-mail. Responding on the issue page itself by clicking on the link in e-mail provides a much better experience and you can attach files easily.

pchtsp:
Thanks @tkralphs I add the files via the web: 
Ok, apparently I cannot attach a file with python extension so here it is in pastebin: https://pastebin.com/783i3Yvm

seangrogan:
Hello!

I made a pull request in line with @stumitchell 's comments

https://github.com/coin-or/pulp/pull/261

pchtsp:
[...]

PR #261, opened 2020-02-05
Inconsistent solver arguments across solver types, quality of life suggestion (My issue #260)

seangrogan:
Hello there,  

So this was a self response to my issue I posted [here](https://github.com/coin-or/pulp/issues/260) (basically figuring out how to do pull requests at the same time).  

Taking guidance from Stu Mitchell's response, I re-wrote the solver initializations that had an explicit parameter that limited the solver's execution time.   

I chose to use the camelCase convention `timeLimit` that seems to be in line with the rest of Pulp.

I included a warning if the user supplies an old parameter (timelimit/maxSeconds) and another warning if the user supplies both the old and new parameter.  I have set up the warnings to take the `timeLimit` parameter if both parameters are passed.  

The glaring omission is the GUROBI_CMD which doesn't have a timeLimit/timelimit/maxSeconds parameter by default.  I didn't want to try and include new parameters in the solver in the same pull request, but I imagine that's a next step?


seangrogan:
I goofed up and used an F-string.  I fixed it and I hope the check will work well for older Python versions.

Sorry about that.

seangrogan:
I used the keyword only arguments (the asterix in a function definition) as a shorthand to separate the future depreciated arguments.

pchtsp:
Thanks for the PR.
Two questions:

1. Isn't there a way to do this without repeating the deprecation code at each time? maybe a common function that is used by all solvers?
2. Could you please do the same with the relative (absolute) gap information?

Thanks again!

seangrogan:
Idea ::  Is there any logic or reason behind not putting these arguments initially in the super-class of the solver? (`class LpSolver`)? 

My idea would be to put some common LP/MIP parameters in the generic solver constructor (Mip gap, absolute gap, time limit, and log file parameters are the only ones that come to mind).

Could we place this in the LPSolver constructor, test if `timelimit` or `maxSeconds` are passed, and test it there?  I'd have to go through and make sure the internal class parameters use the new `timeLimit`, which I initially didn't to make sure.

Are there any other parameters that could fit in here?

pchtsp:
That actually sounds like a very good idea. What do you think @stumitchell 
Another parameter I've used in more than one solver are "threads". Maybe "memory" (limit).

stumitchell:
Sure, placing the code in LpSolver makes sense,

we will have to do solver specific code that will issue a warning if the solver doesn't actually support the keyword.

seangrogan:
Had some free time this afternoon and I coded up CPLEX_CMD - Familiarity plus availability.

I added `timeLimit, mipGap, absMipGap` and it  it seems to work as expected, with warnings for maxSeconds and timelimit.

Y'all don't have any kind of contribution guidelines?  I'm using camelCase as it appears you use it the most often.  The only thing I see no rhyme or reason is argument order

pchtsp:
hello everyone. A lot has changed in the codebase since this PR. But I've added the changes manually on the new files. Check: https://github.com/coin-or/pulp/blob/4ac5e95c628420b990cbb92ed25a15311f38e401/pulp/apis/core.py#L167
for the good parameters in the parent object and 
https://github.com/coin-or/pulp/blob/4ac5e95c628420b990cbb92ed25a15311f38e401/pulp/apis/cplex_api.py#L15 for the example on the cplex api.

Do tell me if you see something weird.

Thanks again.

seangrogan:
@pchtsp thanks for the info

Sorry for the disappearing act.  Got sick and then Canada came to a stop.  Been struggling to get the home office set up.  

I'll have a look at the new code give it a test drive.

pchtsp:
I'm going to close this PR. Not because it was not useful (it was). But we (I) took too long to approve it and we ended up making many changes that broke the merge.
I ended up incorporating the functionality in another commit so it should be already part of the codebase.

Sorry for that. If you think there's still things to add to the master version, feel free to add another PR and I will check it,

PR #263, opened 2020-02-20
changed sense to MIPCL at solving time. We undo the change after solv…

pchtsp:
…ing.


stumitchell:
This was address in https://github.com/coin-or/pulp/pull/264 right

PR #269, opened 2020-03-02
Split solvers

pchtsp:
I reworked the imports to make the explicit everywhere.
I also split all api according to their solver under the same directory.
I also took the chance to corrected some errors, bad formatting and other things pycharm's linter detected on the way.


stumitchell:
Looks good we should get this in master asap before we move into merge hell

PR #271, opened 2020-03-14
Remove the redundant Description variable

rajat-008:
It seems as though after having introduced the long_description variable, the removal of Description was missed upon.


pchtsp:
thanks for the feedback, although is not worth the PR. We will take it out in next changes.

PR #274, opened 2020-03-27
updated pulp and CPLEX_PY to add mip start functionality. Also, impro…

pchtsp:
…ved CPLEX_PY (and pulp) performance with models with many variables.


pchtsp:
What do you think about the changes @stumitchell ? Specially the fact of using `lp._variables` insetad of `lp.variables()` in many places. It does help in performance and I think it has no impact once `lp.variables()` has been executed once.

Issue #276, opened 2020-04-01
Missing code in Transportation Problem Case Study


KCachel:
#### What did you do?
In the[ Transportation Problem Case Study](https://coin-or.github.io/pulp/CaseStudies/a_transportation_problem.html) example it appears that a line of code to make the costs dictionary is missing from the example page.  

#### What did you expect to see?

I think this additional line is needed to make the costs dictionary as described in the text that goes with the example. 
`costs = makeDict([Warehouses, Bars], costs, 0)`

The code as illustrated in the code blocks is missing a dictionary that is referenced later on.

#### What did you see instead?
Notice how in this code block, and subsequent ones the `costs` dictionary is not created.
![image](https://user-images.githubusercontent.com/36267844/78181028-ae9d8600-7431-11ea-99da-2bc642da3343.png)

Useful extra information
-------------------------

The info below often helps, please fill it out if you're able to. :)

#### What operating system are you using?

- [ ] Windows: ( _version:_ ___ )
- [ ] Linux: ( _distro:_ ___ )
- [X] Mac OS: ( _version:_ ___ )
- [ ] Other: ___

#### I'm using python version:

- [ ] 2.7
- [ ] 3.4
- [ ] 3.5
- [X] 3.6
- [ ] Other: ___

#### I installed PuLP via:

- [ ] pypi (pip install pulp)
- [ ] github
- [X ] Other: Anaconda

#### Did you also

- [X] Tried out the latest github version: https://github.com/coin-or/pulp
- [X] Searched for an existing similar issue: https://github.com/coin-or/pulp/issues?utf8=%E2%9C%93&q=is%3Aissue%20

pchtsp:
While checking this problem I realized that this example is the only one that is not referenced to a file in the `examples/` directory.

For example the Sudoku example in `doc/source/CaseStudies/a_sudoku_problem.rst` does this:

```
The introductory commenting and import statement are entered

.. literalinclude:: ../../../examples/Sudoku1.py
    :lines: 1-10

In the unique case of the sudoku problem, the row names, column names and variable option values are all the exact same list of numbers from 1 to 9.

.. literalinclude:: ../../../examples/Sudoku1.py
    :lines: 11-13

```

Do you feel like creating a new python file in the `examples/` directory with the complete transportation model and and then editing the `doc/source/CaseStudies/a_transportation_problem.rst` to match the other `CaseStudies` way of working?

That way we can also run that example during tests.

KCachel:
[...]

PR #277, opened 2020-04-02
resolves #275 : edits to a_transportation_problem.rst

### LINKED ISSUE #275: Pulp LpDot behaviour on list of list wrong (not similar to @ or np.dot)
i have two vectors of variables and constants, the length is 2000, I am using lpDot and it is taking a fairly large amount of time 2-3 secs.
I have similar results with np.Dot or @, So the problem is with the pulp variable which is leading such large time. I also tired cvxpy the [cvxpy var...] @ [constants] of same length array takes 100 ms. So why in pulp it takes so large time ? What is solution or tip for speedup

KCachel:
Edit a_transportation_problem.rst to reference BeerDistributionProblem.py to resolve [issue #275](https://github.com/coin-or/pulp/issues/276). This adds code to the transportation case study so that the model works properly.

### DIFF
--- a/doc/source/CaseStudies/a_transportation_problem.rst
+++ b/doc/source/CaseStudies/a_transportation_problem.rst
@@ -241,25 +241,9 @@ This is added into the above model very simply. Since the objective function and
 constraints all operated on the original supply, demand and cost 
 lists/dictionaries, the only changes that must be made to include another demand node are:
 
-.. code-block:: python
+.. literalinclude:: ../../../examples/BeerDistributionProblemWarehouseExtension.py
+    :lines: 11:31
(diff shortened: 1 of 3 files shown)


pchtsp:
Hello! thanks again. RE 1, I was referring to the code blocks that start in lines 244 and 286.

KCachel:
Hello, I added two new files `BeerDistributionProblemWarehouseExtension.py` and `BeerDistributionProblemCompetitorExtension.py` to the /examples folder to handle the extensions on the case study. Then updated the code in the `a_transportation_problem.rst` to reference those files and provide links to them.

PR #291, opened 2020-05-15
model import and export using dictionaries and json. With tests.

pchtsp:


pchtsp:
Hello Jeremy, thanks for your message. I won't be able to join for tomorrow's call but I could join one day for a chat (the email address was removed by github I think so I didn't get it). 
With regards to the slack channel: it's private and I think I need an invitation to join. Please send one to pchtsp@gmail.com.

thanks again!

PR #293, opened 2020-05-24
Added error check

charlesfox97:
Sometimes the .sol file will say optimal in the first line, but leave constraints unsatisfied.  I think this occurs when the initial values are already optimal.

### DIFF
--- a/pulp/apis/coin_api.py
+++ b/pulp/apis/coin_api.py
@@ -184,6 +184,10 @@ class COIN_CMD(LpSolver_CMD):
                 l = l.split()
                 #incase the solution is infeasible
                 if l[0] == '**':
+                    if status == 1:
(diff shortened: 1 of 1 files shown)


pchtsp:
Hello, thanks for the help. 
Can you please provide a unit test where this problem occurs and is solved by the PR?

charlesfox97:
I'm not sure how to do that correctly.  The example that drove me to suggest this change is: 

https://coin-or.github.io/pulp/guides/how_to_mip_start.html

pchtsp:
I just checked what you say. The problem is apparently within CBC. It does not report the optimal solution when given the input file. I've tried with a recent version of CBC and it works correctly. I will create an issue to update the CBC version.

PR #306, opened 2020-06-27
docs(INSTALL): Update INSTALL

JulianATA:
Hi, guys:
I manually install PuLP under my Linux environment:
```bash
$ uname -a
Linux 5.3.0-61-generic #55~18.04.1-Ubuntu SMP Mon Jun 22 16:40:20 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
```

Find out the testing guide in INSTALL document is outdated:
```python
>>> import pulp
>>> pulp.pulpTestAll()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: module 'pulp' has no attribute 'pulpTestAll'
```

Base on the discussion of [#253](https://github.com/coin-or/pulp/issues/253#issuecomment-563928589), the testing method on Linux now should be:
```bash
$ python3 pulp/tests/run_tests.py
```

Hence I updated the INSTALL documentation.
Hope this is helping!

PR #307, opened 2020-06-27
fix(tests): Import tests correctly for pulpTestAll method

JulianATA:
Hello!
Based on the PR #306 , we should fix the situation:
```python3
>>> import pulp
>>> pulp.pulpTestAll()
AttributeError: module 'pulp' has no attribute 'pulpTestAll'
```

The reason it failed is based on 2 reasons:
* pulp/tests/\_\_init\_\_.py didn't import the run_tests module
* pulp/\_\_init\_\_.py didn't import the tests package

The modules doesn't share the same name with their packages are [submodules](https://docs.python.org/3/reference/import.html#submodules).
We should import submodules in \_\_init\_\_.py.
* The run_tests module does not share the name `tests` with its package.

So, I got 2 solutions here:
1. Import submodules in pulp/tests/\_\_init\_\_.py:
    * Which is provided in this PR.
    * Consistent with the pulp/\_\_init\_\_.py
    * Easy and less modified.
2. Rename the run_tests.py to tests.py:
    * Still need to import tests in pulp/\_\_init\_\_.py

### DIFF
--- a/pulp/__init__.py
+++ b/pulp/__init__.py
@@ -36,6 +36,7 @@ from amply import *
 from .apis import *
 from .utilities import *
 from .constants import *
+from .tests import pulpTestAll
 
 __doc__ = pulp.__doc__
 __version__ = VERSION
--- a/pulp/tests/__init__.py
+++ b/pulp/tests/__init__.py
@@ -1,0 +1,1 @@
+from .run_tests import pulpTestAll
--- a/pulp/tests/test_pulp.py
+++ b/pulp/tests/test_pulp.py
@@ -607,6 +607,12 @@ class PuLPTest(unittest.TestCase):
         else:
             pulpTestCheck(prob, self.solver, [const.LpStatusUnbounded])
 
+    def test_pulpTestAll(self):
+        """
+        Test the availability of the function pulpTestAll
+        """
+        from pulp import pulpTestAll
+        print("\t Testing the availability of the function pulpTestAll")
 
 def pulpTestCheck(prob, solver, okstatus, sol=None,
                   reducedcosts=None,
--- a/pulp/tests/test_pulp.py
+++ b/pulp/tests/test_pulp.py
@@ -651,8 +651,8 @@ class PuLPTest(unittest.TestCase):
         """
         Test the availability of the function pulpTestAll
         """
-        from pulp import pulpTestAll
         print("\t Testing the availability of the function pulpTestAll")
+        from pulp import pulpTestAll
 
     def test_export_dict_LP(self):
         prob = LpProblem("test_export_dict_LP", const.LpMinimize)


pulp/__init__.py
```diff
@@ -36,6 +36,7 @@
 from .apis import *
 from .utilities import *
 from .constants import *
+from .tests import *
```
pchtsp, line 39: I think I prefer only exposing `pulpTestAll` just to be safe so:

`from .tests import pulpTestAll`

pulp/tests/__init__.py
```diff
@@ -0,0 +1 @@
+from .run_tests import *
```
pchtsp, line 1: while you're at it, add a new line here.

pulp/__init__.py
JulianATA: You are right. It is better to expose less members of package.

pulp/tests/__init__.py
JulianATA: Same, modified to import only pulpTestAll.

JulianATA:
The most easy to intuitive way to check:
```python
>>> import pulp
>>> pulp.pulpTestAll()
AttributeError: module 'pulp' has no attribute 'pulpTestAll'
```
It'll return error while calling `pulp.pulpTestAll()` before.

```python
>>> import pulp
>>> pulp.pulpTestAll()
Solver <class 'pulp.apis.coin_api.PULP_CBC_CMD'> available
Solver <class 'pulp.apis.cplex_api.CPLEX_DLL'> unavailable
Solver <class 'pulp.apis.cplex_api.CPLEX_CMD'> unavailable
.
.
.
Ran 130 tests in 13.370s

OK
```
It'll log the testing messages after.

* For creating an unit test for it, we could use the built-in [`hasattr()`](https://docs.python.org/3/library/functions.html#hasattr) function.

> hasattr(object, name)¶
>> The arguments are an object and a string. The result is True if the string is the name of one of the object’s attributes, False if not. (This is implemented by calling getattr(object, name) and seeing whether it raises an AttributeError or not.)

The unit test created by hasattr():
```python
import pulp
assert(hasattr(pulp,'pulpTestAll'))
```
It'll log out assertion error before:
```bash
$ python3 test.py
  File "test.py", line 2, in <module>
   assert(hasattr(pulp,'pulpTestAll'))
AssertionError
```
It'll pass without any message after.
```bash
$ python3 test.py
```
Thanks for your checking!

pchtsp:
Thanks again. You're right: for some reason I though that in order to run that kind of test we would need to have `pulp` installed in the environment, but it's not necessary apparently.

I propose you add a new unit test in `tests/test_pulp.py` like so:

```python
def test_pulpTestAll(self):
    """
    Test the availability of the function pulpTestAll
    """
    from pulp import pulpTestAll
```

I can confirm it fails without the changes and it runs ok with the changes.

JulianATA:
@pchtsp , this is a very clean way to confirm! The proposed test is added.
Thanks!

JulianATA:
Conflict resolved!

PR #308, opened 2020-06-29
fixes #215

### LINKED ISSUE #215: use TMP file in COIN_CMD
Hello Stu,

I just came across the fact that the CBC solver does not use the TMP file because of this comment. https://github.com/coin-or/pulp/blob/183dbdff3f6a4e559af8d503752c27c95b0c2a00/src/pulp/solvers.py#L1368

I did the following:

1. Deleted the following two lines.
2. Tested solving using CBC (in windows).
3. Checked that it correctly uses the tmpdir.

What I used for the test:

1. Windows 10.
1. pulp's CBC (it says version 2.9.0).
2. python 3.6.

Should we just take out this three lines? Or is there some other test I should be doing?

pchtsp:


tkralphs:
Just FYI, when I activated the CLA Assistant across all of COIN-OR, I think it deleted the one I had set up for PuLP initially, so you'll have to sign again, but this should be the last time. Sorry about that!

PR #309, opened 2020-06-29
Fix cplex dll

pchtsp:
Hello @stumitchell ,

I've been trying to fix the issues that were appearing with the CPLEX_DLL api, at least in my ubuntu machine with python3.6. I'm currently stuck at running the `solver_obj.copy()` method when created by CPLEX_DLL. It crashes when copying the object without any error message. Any idea on why this would be the case? I use the copying during the test suite.


stumitchell:
I have very little to offer here.

My suggestion would be to remove this solver and only allow CPLEX_py and CPLEX_CMD

pchtsp:
too many commits. I create a new one.

PR #311, opened 2020-07-02
Validate status codes before assignment

JulianATA:
Hi guys, 
I'm recently studying about codes in PuLP.
Found the` TODO: check if status are valid status codes.`
This modifications should have added this feature!
For the testing it, I added a test for it in pulp/tests.

### DIFF
--- a/pulp/pulp.py
+++ b/pulp/pulp.py
@@ -1901,7 +1901,12 @@ class LpProblem(object):
         :param sol_status: code for the status of the solution
         :return:
         """
-        # TODO: check if status are valid status codes
+        if status not in const.LpStatus:
+            raise const.PulpError("Invalid status code: "+str(status))
+
+        if sol_status is not None and sol_status not in const.LpSolution:
+            raise const.PulpError("Invalid solution status code: "+str(sol_status))
+
         self.status = status
         if sol_status is None:
             sol_status = const.LpStatusToSolution.get(status, const.LpSolutionNoSolutionFound)
--- a/pulp/tests/test_pulp.py
+++ b/pulp/tests/test_pulp.py
@@ -787,7 +787,13 @@ class PuLPTest(unittest.TestCase):
         # CHOCO has issues when given a time limit
         if self.solver.name != 'PULP_CHOCO_CMD':
             pulpTestCheck(prob, self.solver, [const.LpStatusOptimal], {x: 4, y: -1, z: 6, w: 0})
-
+    
+    def test_assignInvalidStatus(self):
+        print("\t Testing invalid status")
+        t = LpProblem('test')
+        Invalid = -100
+        self.assertRaises(const.PulpError, lambda: t.assignStatus(Invalid))
+        self.assertRaises(const.PulpError, lambda: t.assignStatus(0, Invalid))
 
 def pulpTestCheck(prob, solver, okstatus, sol=None,
                   reducedcosts=None,


pulp/pulp.py
```diff
@@ -1901,7 +1901,14 @@ def assignStatus(self, status, sol_status=None):
         :param sol_status: code for the status of the solution
         :return:
         """
-        # TODO: check if status are valid status codes
+        if status not in const.LpStatus:
```
pchtsp, line 1904: I don't think that printing a message is a good idea. I would raise an error or a warning. see the examples with warnings `warnings.warn("Overwriting previously set objective.")` or errors `raise const.PulpError("Duplicated names found in variables:\nto export the model, variable names need to be unique")` in this same file. What do you think?

pulp/tests/test_pulp.py
```diff
@@ -787,7 +787,16 @@ def test_timeLimit(self):
         # CHOCO has issues when given a time limit
         if self.solver.name != 'PULP_CHOCO_CMD':
             pulpTestCheck(prob, self.solver, [const.LpStatusOptimal], {x: 4, y: -1, z: 6, w: 0})
-
+    
+    def test_assignStatus(self):
```
pchtsp, line 791: Most examples already test that `const.LpStatus` is filled correctly but they do not check `const.LpSolution`. Check the function just below `pulpTestCheck`. Maybe we can make that function test that for solution status. Also, it would be interesting to add tests were we add an incorrect status and raise the error or warning.

pulp/pulp.py
JulianATA: I thought I'll modify to `const.PulpError()`, since it is more like an error than a dangerous behavior. Thanks for your comments!

pulp/tests/test_pulp.py
JulianATA: I didn't notice that `const.LpStatus` has been tested, cause there is no directly `assignStatus()` test. For this PR, I thought invalid status check is a good idea.
The check for the validity of solution status would be easy, but the verification of solution status will be a massive modification.(Since it requires to know all the solution status of examples.)
I'll go for invalid status test first!

pulp/tests/test_pulp.py
```diff
@@ -787,6 +787,24 @@ def test_timeLimit(self):
         # CHOCO has issues when given a time limit
         if self.solver.name != 'PULP_CHOCO_CMD':
             pulpTestCheck(prob, self.solver, [const.LpStatusOptimal], {x: 4, y: -1, z: 6, w: 0})
+    
+    def test_assignInvalidStatus(self):
+        print("\t Testing invalid status")
+        t = LpProblem('test')
+        Invalid = -100
+        try:
```
pchtsp, line 795: One last thing, could you use the `self.assertRaises` here? I think is cleaner. Check this recent example: https://github.com/coin-or/pulp/blob/3414dc943e2985e0b20158a0850b9771e62fd9b6/pulp/tests/test_pulp.py#L846

JulianATA: Thanks for introducing this method to me, the test has been modified.

pchtsp:
Cool! You just need to resolve the conflict by merging the master into your branch correctly and then do a last push so I can merge this into the master.

JulianATA:
Thanks, the conflict has been resolved!

PR #321, opened 2020-07-16
Fix makeDict's behavior

matheusMoreno:
Change the condition on the recursive function `__makeDict` so that a defaultdict is returned when `default` is specified.

Also, add a test `test_makeDict` on `pulp/tests/test_pulp.py` that creates two dictionaries with the same headers and values, but one with a default value and the other without it. The test passes if `dict_with_default` returns the default value when we try to access a non-existent key and if `dict_without_default` raises a KeyError when we try to access the same key.

### DIFF
--- a/pulp/tests/test_pulp.py
+++ b/pulp/tests/test_pulp.py
@@ -819,7 +819,7 @@ class PuLPTest(unittest.TestCase):
 
     def test_makeDict_behavior(self):
         """
-        Test if the makeDict function is returning the expected value.
+        Test if makeDict is returning the expected value.
         """
         headers = [["A", "B"], ["C", "D"]]
         values = [[1, 2], [3, 4]]
@@ -832,7 +832,7 @@ class PuLPTest(unittest.TestCase):
 
     def test_makeDict_default_value(self):
         """
-        Test if the makeDict function is returning a defaultdict when default is specified.
+        Test if makeDict is returning a default value when specified.
         """
         headers = [["A", "B"], ["C", "D"]]
         values = [[1, 2], [3, 4]]
--- a/pulp/tests/test_pulp.py
+++ b/pulp/tests/test_pulp.py
@@ -5,6 +5,7 @@ from pulp.constants import PulpError
 from pulp.apis import *
 from pulp import LpVariable, LpProblem, lpSum, LpConstraintVar, LpFractionConstraint
 from pulp import constants as const
+from pulp.utilities import makeDict
 import unittest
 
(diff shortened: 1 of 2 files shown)


pchtsp:
Thank you very much for the test. One last thing, could you add another test ?

I'd like to have a test for the "expected" behavior of makeDict too (not testing the defaults but an assert on the whole contents of the resulting dictionary). I'd do it myself but I don't really know how this function is used. 

Thanks again!

F.

matheusMoreno:
Done! Let me know what you think.

pulp/utilities.py
```diff
@@ -156,7 +156,7 @@ def __makeDict(headers, array, default=None):
     else:
         for i, h in enumerate(headers[0]):
             result[h], defaultvalue = __makeDict(headers[1:], array[i], default)
-    if default is None:
+    if not default is None:
```
pchtsp, line 159: I find `if default is not None` easier to read. And it's still valid python.

pulp/tests/test_pulp.py
```diff
@@ -816,6 +817,37 @@ def test_logPath(self):
...
+        Test if the makeDict function is returning a defaultdict when default is specified.
+        """
+        headers = [["A", "B"], ["C", "D"]]
+        values = [[1, 2], [3, 4]]
+        dict_with_default = makeDict(headers, values, default=0)
+        dict_without_default = makeDict(headers, values)
+
+        print("\t Testing makeDict default value behavior")
+        # Check if a default value is passed, and if a KeyError is raised
+        assert dict_with_default["X"]["Y"] == 0
```
pchtsp, line 844: just to be consistent with the `unitttest` functions, you could uses `self.assertEqual(dict_with_default['X']['Y'], 0)`.

pulp/tests/test_pulp.py
```diff
@@ -816,6 +817,37 @@ def test_logPath(self):
...
+        try:
+            z = dict_without_default["X"]["Y"]
+            raise PulpError("Test for makeDict default value failed")
```
pchtsp, line 847: finally, I think `self.failUnlessRaises` can be useful here:
```
        _func = lambda: dict_without_default["X"]["Y"]
        self.failUnlessRaises(KeyError, _func)
```

matheusMoreno: Done! I used `self.assertRaises` because unittest raised a `DeprecationWarning` when I called `self.failUnlessRaises`.

PR #322, opened 2020-07-21
Update how_to_configure_solvers.rst

rneu31:
Fix typo while reading through the documentation


pchtsp:
Hello! Thanks for the PR. Just so it "justifies" the commit, could you check if there are other issues in the docs? I think I've seen some errors in https://coin-or.github.io/pulp/guides/how_to_export_models.html. In the same file you mention, there are other issues I think (e.g., "Any arguments **passes** to this function are passed to the constructor").

PR #323, opened 2020-07-26
Update wedding.py

bmedboen:
Suggestion to use lpSum instead of sum to make code more efficient and aligned with best practice and the other cases. See https://stackoverflow.com/questions/63081381/when-to-use-sum-v-s-lpsum-using-pulp/63081480#63081480 for background. 

The code runs significantly faster.

Note that the import statement varies among the different pulp cases. In this example and the "A Two Stage Production Planning Problem" the statement used is "import pulp". 

Also note that this is my first PR.


pchtsp:
Thanks for the PR! Could you please edit the wedding_initial.py file with
the same changes?

Also, at the end of the following file there is a copied version of that
problem:
doc/source/guides/how_to_export_models.rst

Could you change that too?

Thanks again for your help!

F.

PR #324, opened 2020-07-27
Update wedding_initial.py

bmedboen:
Code is changed to use pulp.lpSum intead of Sum according to PR 323. I am unsure how to merge PR, but I will try.


pchtsp:
No need to create a new PR. You just need to add the two new commits to the previous branch (patch1) and push that to the original branch to github. This will update the old PR.

PR #325, opened 2020-07-28
fix(apis/gurobi_api): Replace set parameter `_test` to system command

JulianATA:
Hi!
Recently I have done some experiments on pulp using solver from gurobi.
While importing gurobi, pulp will set the parameter `_test` to 0.
However the parameter `_test` might be outdated, it'll show the warning below:
```python3
>>> import pulp
Using license file /path/to/gurobi.lic
Academic license - for non-commercial use only
No parameters matching '_test' found
```
Removing the line: `gurobipy.setParam("_test", 0)` is able to fix this redundant warning.
```python3
>>> import pulp
```

I don't know if there is a parameter for test in latest version of gurobi.
The parameters in [doc](https://www.gurobi.com/documentation/9.0/refman/parameters.html) of gurobi are mainly about the solving process.

Thanks for your review again!


pchtsp:
Hello!

The parameter has never existed. I just put that there to test the library
at that moment and force an error if there is no license. I did no find a
method to ask if there is a valid license. Maybe there is one?

An alternative could be to replace that parameter with some other one that
has no impact. In fact, the idea is only to test the library in some way.

JulianATA:
@pchtsp, thanks for your clear explanation!
I thought the closest way to perform these 2 task(checking api and license) is using the system method in gurobipy.
The system method can use gurobi_cl to check if the gurobi license exist.

Thanks for your review, again!

JulianATA:
After this modification it'll like:
```python
>>> import pulp
Using license file /path/to/gurobi.lic
Set parameter LogFile to value gurobi.log
Academic license - for non-commercial use only
```

pchtsp:
thanks. I have two questions:

1. Does this change throw an exception like the previous code when there is no gurobi license or the license is expired? I'm assuming it does but we should confirm it. Aren't we assuming in this case that `gurobi_cl` is in the PATH? It may not be? I'm not sure.
2. Could we use something like that (i.e., running `gurobi_cl --license`) to test in the GUROBI_CMD solver if there is a valid license when deciding if we give access to the API or not? Currently if `gurobi_cl` exists but there is not valid license, the GUROBI_CMD api throws an exception when trying to solve.

JulianATA:
1. First, it does not throw back an exception when there is no license. Since, `gurobipy.system()` does only execute the command.
Instead, I found a command does test the license without an warning:`gurobipy.gurobi._getdefaultenv()`.
With this method, we can get rid of using `gurobi_cl`.

JulianATA:
@pchtsp,
1. For the first question, it actually does not throw back an exception when there is no valid license. Instead, I found another command to replace the original command which does not rely on `gurobi_cl`.
It is `gurobipy.gurobi._getdefaultenv()`. which will raise an error while there is no valid license.

JulianATA:
2. Yeah, I think it may help to test if `gurobi_cl --license` or `gurobipy.gurobi._getdefaultenv()` may help.
The testing result below is without a valid license of gurobi while running `pulptest`:
```bash
Solver <class 'pulp.apis.gurobi_api.GUROBI'> unavailable
Solver <class 'pulp.apis.gurobi_api.GUROBI_CMD'> available
```
Under this condition, the test will still test for gurobi to solve the linear problem, which will leads to failure of the test.
I guess the gurobi api should be available only if both GUROBI and GUROBI_CMD are available.

Issue #150, opened 2017-09-22
Tests for SCIP 


bbalegere:
Currently, tests for SCIP are not present. I added SCIP_CMD to pulpTestAll, but the test failed with the below error. Does this mean that Pulp does not fully support SCIP ? Is there new buildout required ?

Solver pulp.solvers.YAPOSIB unavailable
         Testing zero subtraction
         Testing inconsistant lp solution
Can't read SCIP solver output: 'no solution available\n'
('* Solver', <class pulp.solvers.SCIP_CMD at 0x0000000002E928E8>, 'failed.')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "d:\Python27\lib\site-packages\pulp\pulp.py", line 2311, in pulpTestAll
    raise PulpError("Tests Failed")

pchtsp:
We need someone that has SCIP installed to test it and correct it.

pchtsp:
closed by https://github.com/coin-or/pulp/pull/332

PR #332, opened 2020-08-31
Issue 150/scip tests

rimaddo:
# Description
- Fixes tests for SCIP when it is included as a solver as requested in https://github.com/coin-or/pulp/issues/150
- Adds handling for when SCIP cannot find a solution (this was where most of the tests were failing)
- Correctly categorises status in failing cases. Most respond with 'infeasible or unbounded' which maps to the status `constants.LpStatusNotSolved`.
- I also added some increased descriptions to tests that log `Testing continuous LP solution` (there were 5) so I could tell the difference, let me know if you would prefer this not to be committed and I will remove!

### DIFF
--- a/pulp/apis/scip_api.py
+++ b/pulp/apis/scip_api.py
@@ -62,6 +62,11 @@ class SCIP_CMD(LpSolver_CMD):
         'unbounded': constants.LpStatusUnbounded,
         'infeasible or unbounded': constants.LpStatusNotSolved,
     }
+    NO_SOLUTION_STATUSES = {
+        constants.LpStatusInfeasible,
+        constants.LpStatusUnbounded,
+        constants.LpStatusNotSolved,
+    }
 
     def defaultPath(self):
         return self.executableExtension(scip_path)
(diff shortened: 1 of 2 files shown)


pulp/apis/scip_api.py
```diff
@@ -108,35 +113,38 @@ def actualSolve(self, lp):
...
-            # Look for an objective value. If we can't find one, stop.
-            try:
-                line = f.readline()
-                comps = line.split(': ')
-                assert comps[0] == 'objective value'
-                assert len(comps) == 2
-                float(comps[1].strip())
-            except:
-                raise PulpSolverError("Can't read SCIP solver output: %r" % line)
+            if status not in SCIP_CMD.NO_SOLUTION_STATUSES:
```
pchtsp, line 129: maybe better to flip this to  and exit early:

```python
if status in SCIP_CMD.NO_SOLUTION_STATUSES:
    return status, values
```

pulp/tests/test_pulp.py
```diff
@@ -149,7 +149,7 @@ def test_pulp_012(self):
             # CPLEX_DLL Does not report unbounded problems, correctly
             # GUROBI_CMD has a very simple interface
             pulpTestCheck(prob, self.solver, [const.LpStatusNotSolved])
-        elif self.solver.__class__ in [PULP_CHOCO_CMD, CHOCO_CMD]:
+        elif self.solver.__class__ in [PULP_CHOCO_CMD, CHOCO_CMD, SCIP_CMD]:
```
pchtsp, line 152: wouldn't SCIP_CMD return `const.LpStatusNotSolved` ? if this is the case, better to put it with `CPLEX_DLL` and `GUROBI_CMD` above. Or on some other if clause that matches SCIP's output status.

pulp/tests/test_pulp.py
```diff
@@ -356,7 +357,7 @@ def test_pulp_030(self):
         prob += -y + z == 7.5, "c3"
         self.solver.mip = 0
         print("\t Testing MIP relaxation")
-        if self.solver.__class__ in [GUROBI_CMD, PULP_CHOCO_CMD, CHOCO_CMD, MIPCL_CMD]:
+        if self.solver.__class__ in [GUROBI_CMD, PULP_CHOCO_CMD, CHOCO_CMD, MIPCL_CMD, SCIP_CMD]:
```
pchtsp, line 360: SCIP has no option to relax the problem ? or is it just not part of the current interface? I ask in order to add it as a TODO somewhere in case it's just missing from the SCIP_CMD interface.

pulp/tests/test_pulp.py
```diff
@@ -648,7 +649,7 @@ def test_pulp_123(self):
         elif self.solver.__class__ in [CPLEX_DLL, GUROBI_CMD]:
             # GLPK_CMD Does not report unbounded problems, correctly
             pulpTestCheck(prob, self.solver, [const.LpStatusNotSolved])
-        elif self.solver.__class__ in [PULP_CHOCO_CMD, CHOCO_CMD]:
+        elif self.solver.__class__ in [PULP_CHOCO_CMD, CHOCO_CMD, SCIP_CMD]:
```
pchtsp, line 652: here too, can we match the actual status from SCIP from one of the above if clauses?

pchtsp (commented):
Thank you for your contribution! 

I've added some minor comments / questions.

I do not have SCIP so I cannot test this easily. Do you know if there is a way to install SCIP easily as a command line (from a Linux repository, for example?). This way I could try it myself but, more importantly, we could also add it to the automatic unittests that are executed in github.

Thanks again.

pulp/apis/scip_api.py
rimaddo: Nicer I agree, will change!

pulp/tests/test_pulp.py
rimaddo: Yes my mistake

pulp/tests/test_pulp.py
rimaddo: I think you can relax problems with SCIP by adding a relaxation handler? However, the current interface here doesn't seem to be handling it. Though I don't know much about this in general so I could be wrong.

pulp/tests/test_pulp.py
rimaddo: Sorry I think i misread this and the other one, they both should have been in the group above, categorised now.

rimaddo:
> Thank you for your contribution!
> 
> I've added some minor comments / questions.
> 
> I do not have SCIP so I cannot test this easily. Do you know if there is a way to install SCIP easily as a command line (from a Linux repository, for example?). This way I could try it myself but, more importantly, we could also add it to the automatic unittests that are executed in github.
> 
> Thanks again.

No worries, thanks for the comments!

You can [download SCIP here](https://www.scipopt.org/index.php#download) however it has a bit of a manual step where you need to fill in a form ie its not set-up for downloading automatically within integration tests. I don't know another way to install the command line interface off the top of my head, by I can try to do a bit of digging on this perhaps?

PR #334, opened 2020-09-02
IPython printing fix for CBC+GUROBI solvers api (coin-or#333)

AlfaimaSB:
If I try to print log with GUROBI_CMD(msg=1) function it will not work when using an IPython environment like Jupyter Notebook or Spyder. It works perfectly if the script is run from the terminal but I need Spyder cells to teach. I can use %%python cell magic but then I would have to run all the script of an optimization model in a cell, which is not the best workaround when you need to teach with the least cognitive load as possible (not opening the console in the middle of the class). I only changed the default pipe: None for subprocess.PIPE to capture the terminal output and then I print it in real time. I am of course not the best programmer but I hope this helps. I leave the compatibility to you, thanks!!! 😊✨

### DIFF
--- a/pulp/apis/coin_api.py
+++ b/pulp/apis/coin_api.py
@@ -136,8 +136,9 @@ class COIN_CMD(LpSolver_CMD):
             cmds += "initialSolve "
         cmds += "printingOptions all "
         cmds += "solution "+tmpSol+" "
+
         if self.msg:
-            pipe = None
+            pipe = subprocess.PIPE #This will allow the output to consider the things also being printed in the terminal by the subprocess
         else:
             pipe = open(os.devnull, 'w')
         logPath = self.optionsDict.get('logPath')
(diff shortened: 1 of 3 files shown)


pchtsp:
Thanks for the PR. I'm not sure the approach is the best one, though.  The polling of the subprocess and seems it could make things messy, I really do not know enough to judge. Have you tried this with larger problems?

Isn't there a more straightforward way to that? Being such a specific use-case, I'm thinking that maybe we should offer way to configure the piping of the output easily so this case can be configured from outside pulp without the need to edit the source code. Would providing a pipe argument to the solve function help? Is there a way to do something along these lines?

I would not want to change this important part of the solving process with CBC until I'm sure it's going to work for all cases.

thanks again!

pchtsp:
too much time with no response. I'll close the PR.

PR #336, opened 2020-09-12
Example in documentation, b to a2 to avoid glpk error symbol 'b' in wrong position 

juanpabloaj:
If I followed the original example

```
import pulp as pl
model = pl.LpProblem("Example", pl.LpMinimize)
solver = pl.GLPK_CMD()
_var = pl.LpVariable('a')
_var2 = pl.LpVariable('b')
model += _var + _var2 == 1
result = model.solve(solver)
```

and when I used glpk I got this error 

```
51e9991f96fa4fca84d1c57df2ba6fd5-pulp.lp:9: symbol 'b' in wrong position
CPLEX LP file processing error
  File "first.py", line 7, in <module>
    result = model.solve(solver)
    ..., line 93, in actualSolve
    raise PulpSolverError("PuLP: Error while executing "+self.path)
pulp.apis.core.PulpSolverError: PuLP: Error while executing glpsol
```

I found the explanation in this mail list.

> I did read it before posting and in my version (4.11) it is made
> explicit that  "[B|b]ounds" and "bound" are synonymous

https://lists.gnu.org/archive/html/help-glpk/2007-04/msg00026.html

After I changed the letter `b` to another one, I could use the example.

### DIFF
--- a/doc/source/guides/how_to_configure_solvers.rst
+++ b/doc/source/guides/how_to_configure_solvers.rst
@@ -72,7 +72,7 @@ Imagine using the ``CPLEX_CMD`` solver, the first one is really simple:
     model = pl.LpProblem("Example", pl.LpMinimize)
     solver = pl.CPLEX_CMD(path=path_to_cplex)
     _var = pl.LpVariable('a')
-    _var2 = pl.LpVariable('b')
+    _var2 = pl.LpVariable('a2')
     model += _var + _var2 == 1 
     result = model.solve(solver)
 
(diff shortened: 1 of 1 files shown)


pchtsp:
Thanks for the PR. I do think that more needs to be done. Maybe adding an assertion to check for keyword names before solving an throwing a better message from pulp. That's why I prefer if you create an issue first. If you want to do those changes too, feel free to push another commit with the necessary changes. If not, we will have to create a new PR.

PR #344, opened 2020-09-30
raise TypeError when adding or subtracting a string.

koreiklein:
The current code generates an infinite loop when you try to add or subtract a string from an LpAffineExpression.  This case seems like a common enough failure mode that it'll be good to throw a more user-friendly error.


pchtsp:
Thanks for the PR. Could you please add a unit test to catch this ? That way we're sure it does not happen again.

pchtsp:
Feel free to re-open the PR with the tests. Or add an issue for someone else to do it.

PR #349, opened 2020-10-20
Read mps

pchtsp:


dakenblack:
Is there a discussion around merging this? When would this be merged?

pchtsp:
Good day,

There are no issues with this PR, it's just that the code has not been tested exhaustively. I was waiting on using these functions more in some application before merging. From what I've tested, it works correctly. 

I'm going to merge it only because it's been some time and I don't want to keep the PR forever. Use it at your own risk though!

PR #407, opened 2021-02-20
Include sol status in valid check

rimaddo:
- As discussed in https://github.com/coin-or/pulp/pull/399 [here](https://github.com/coin-or/pulp/pull/399#issuecomment-775508325) this is making the change such that unbounded problems (and also infeasible / ones without a solution depending on how the solver categorises them) are definitely categorised as "invalid", even if the variables and constraints are valid.
- Not sure if you want this, no worries if not! But it made sense to me for completeness so thought I would suggest.

Many thanks


pchtsp:
I'm not sure about this change. What happens in the following cases?:

**case 1**:
1. user solves a valid problem and gets an "no solution found". Does this mean the problem is invalid? The fact the solver did not find a solution does not mean that the problem "is invalid". Even a valid problem can be "infeasible".

**case 2**:

1. user solves an invalid problem, gets a "infeasible status".
2. the user edits the model and "makes it valid".
3. the user executes the `valid` method. It returns `false` even if the model *is* valid. Because the model retains the "old status" from the previous solving.

rimaddo:
> I'm not sure about this change. What happens in the following cases?:
> 
> **case 1**:
> 
> 1. user solves a valid problem and gets an "no solution found". Does this mean the problem is invalid? The fact the solver did not find a solution does not mean that the problem "is invalid". Even a valid problem can be "infeasible".
> 
> **case 2**:
> 
> 1. user solves an invalid problem, gets a "infeasible status".
> 2. the user edits the model and "makes it valid".
> 3. the user executes the `valid` method. It returns `false` even if the model _is_ valid. Because the model retains the "old status" from the previous solving.

Right, I can see your point in case 1 that `LpSolutionNoSolutionFound` probably doesn't belong in this group because it doesn't indicate necessarily that the problem is invalid, if it's only been given a very short time for example.

For 2 I guess you would need to have something where if a model is "edited" you reset it's `status` to `const.LpStatusNotSolved` (and check that instead of `sol_status`)?

Would these changes help or more generally do you thing adding a `sol_status` check to `valid` is not needed?

Thanks!

pchtsp:
Again, thanks for your help.
I think it's better to leave it as it is in the master: that way the `valid` method will continue to exclusively concern the variables and constraints needed to build the model (i.e., the "input" for the solver). If a user wants to get information on the status of the solution, he or she would have to use a different method that checks the status of problem or solution (i.e., the "output" from the solver).

More documentation on that `valid` method would definitively help make this understood (I see that many of adjacent methods also lack doc strings). Feel free to create an issue  for / contribute on fixing the documentation :  )

rimaddo:
> Again, thanks for your help.
> I think it's better to leave it as it is in the master: that way the `valid` method will continue to exclusively concern the variables and constraints needed to build the model (i.e., the "input" for the solver). If a user wants to get information on the status of the solution, he or she would have to use a different method that checks the status of problem or solution (i.e., the "output" from the solver).
> 
> More documentation on that `valid` method would definitively help make this understood (I see that many of adjacent methods also lack doc strings). Feel free to create an issue for / contribute on fixing the documentation : )

No worries, I will close this then and try and look at the documentation when I next have some time. Thanks

Issue #83, opened 2015-05-28
pulp silently ignores 'double constraints'


senarclens:
In Python, it is possible to test for a variable to be within a range (ie. 1 <= a <= 5). Hence, this syntax is also legal in pulp constraints - however the first part is silently ignored.

I recommend one of three resolutions:
a) allow the constraint to directly work as entered
b) automatically split the constraint set in two sets - one for the first part and one for the second
c) keep the current behavior but at least issue a warning

Test case:
import pulp
a = pulp.LpVariable('a', lowBound=0, cat=pulp.LpContinuous)
milp = pulp.LpProblem("TC", pulp.LpMinimize)
milp += a
milp += 1 <= a <= 5
print(milp)

Expected behavior:
a >= 1 and hence the optimal solution is a = 1 OR a warning is issued that "1 <=" will be ignored

Actual behavior:
First part of constraint `milp += 1 <= a <= 5` (1 <= a) is silently ignored.

dassouki:
This to me sounds more like a feature request than a bug. However, the fact that it's not spitting out an error message is a bug in my opinion

stumitchell:
yeah looks like a bug to me
The warning should at least be added

lsorber:
+1 Would be nice to get this fixed.

lleeoo:
To be clear, a workaround for this is to write the expression as two constraints, i.e., to add to the code above: 

```
milp += 1 <= a
```

pchtsp:
I've been trying to solve this but I'm unable to get both values in any of the methods in LpVariable and LpProblem.
From what I've seen the order is as follows in the provided example:

1.` a__ge__.(1)` is executed and returns LpAffineExpression(a) >= 1, which is a constraint (let's call it `c_1`)
2. `a__le__(5)` is executed and returns LpAffineExpression(a) <= 5, which is a constraint (`c_2`)
3. `milp.__iadd__(c_2)` is called and adds the second one.

So, I have no idea how we can detect that there are two values being passed. Is there any internal python method like `__ge__` but for interval comparison?

pchtsp:
So apparently, we can "catch" the error if we stop evaluating affine expressions as a boolean. Making this change breaks many other tests of course since we do evaluate affine expressions as booleans in `pulp`. We could just replace all those references by some method like `LpConstraint.is_filled()` or something along those lines.

```
    def __bool__(self):
        raise const.PulpError('An affine expression has no boolean value.')
[...]

PR #420, opened 2021-03-28
Issue 83.1 pulp doesnt ignore double constraints

rimaddo:
Addresses the issue described in https://github.com/coin-or/pulp/issues/83 where double comparison constraints have their first part ignored and only the second segment of the comparison respected.

As described in https://github.com/coin-or/pulp/issues/83#issuecomment-787116984, this is because currently there is only one version of self for any constraint lp expression, which is not built upon but instead defined over when a second comparison segment is made.

This pull request suggests adding tracking of sub_constraints, so the first part of the comparison is not lost, and both parts can be added to the greater problem.

### DIFF
--- a/pulp/pulp.py
+++ b/pulp/pulp.py
@@ -137,7 +137,6 @@ except ImportError:
 
 import re
 
-
 class LpElement(object):
     """Base class for LpVariable and LpConstraintVar
     """
@@ -145,7 +144,6 @@ class LpElement(object):
     illegal_chars = "-+[] ->/"
     expression = re.compile("[{}]".format(re.escape(illegal_chars)))
     trans = maketrans(illegal_chars, "________")
-
     def setName(self, name):
         if name:
             if self.expression.match(name):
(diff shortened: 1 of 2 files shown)


pchtsp:
Thanks for the PR!
I'm skeptic of adding a temporal "constraint" property to the model object. And also about the other two properties to the Constraint object. I'm guessing it's not possible to implement the change without those? 

It unfortunately looks too hacky to add as is. I would like to have the functionality but only if there is some "simple" way to add it without having to create temporal properties.

Of course, let me know what you think of this. And as as always thanks for the help :  )

rimaddo:
> Thanks for the PR!
> I'm skeptic of adding a temporal "constraint" property to the model object. And also about the other two properties to the Constraint object. I'm guessing it's not possible to implement the change without those?
> 
> It unfortunately looks too hacky to add as is. I would like to have the functionality but only if there is some "simple" way to add it without having to create temporal properties.
> 
> Of course, let me know what you think of this. And as as always thanks for the help : )

No worries, yeah it's a bit of a hard one to fix simply, I couldn't work out a way to do it without adding in some sort of tracking property. Will have a bit more of a think, thanks!

rimaddo:
Closing for now, will make a new pr if I think of a simpler way.

PR #429, opened 2021-04-08
Handling user provided gaps for SCIP

wookenny:
Based on PR #427, this PR proposed to enable `absGap` and `relGap` for the scip api.

**Changes** (on top of #427)
- passing arguments given to solver
- option handling aligned with the one for `coin_api.py`
- status for "gap limit reached" changed to OPTIMAL (optimal wrt. given gap)
-  measuring time corrected (`clock()` does not measure the time for the subprocess call)

**Example**
Bin packing MIP with known optimum (number of bins). 
https://github.com/wookenny/pulp/blob/6f36880525356466e49f0f97af61412a39764384/SCIP_handling_gap_reached.ipynb

To be considered later: 
If a `timeLimit` is given
- `COIN_CMD` solved the given binpacking problem with 11 bins, 10 is optimal. Pulp reports this solution as `OPTIMAL`.
- `SCIP` fins 11 bins as well, the status is "time limit reached". Pulp reports status Not solved.

**At maintainer:** I can also remove the commits from #427 to obtain an independent PR.

### DIFF
--- a/pulp/apis/scip_api.py
+++ b/pulp/apis/scip_api.py
@@ -30,7 +30,6 @@ import os
 import io
 from .. import constants
 import sys
-import resource
 import itertools
 
 
@@ -109,9 +108,9 @@ class SCIP_CMD(LpSolver_CMD):
         stdout = self.firstWithFilenoSupport(sys.stdout, sys.__stdout__)
         stderr = self.firstWithFilenoSupport(sys.stderr, sys.__stderr__)
 
-        self.solution_time = -resource.getrusage(resource.RUSAGE_CHILDREN).ru_utime
+        self.solution_time = -clock()
         subprocess.check_call(proc, stdout=stdout, stderr=stderr)
-        self.solution_time += resource.getrusage(resource.RUSAGE_CHILDREN).ru_utime
+        self.solution_time += clock()
 
         if not os.path.exists(tmpSol):
             raise PulpSolverError("PuLP: Error while executing "+self.path)
--- a/pulp/apis/scip_api.py
+++ b/pulp/apis/scip_api.py
@@ -99,8 +99,8 @@ class SCIP_CMD(LpSolver_CMD):
             proc.append('-q')
         proc.extend(['-c', 'optimize', '-c', 'write solution "%s"' % tmpSol, '-c', 'quit'])
 
-        stdout = self.none_if_fileno_unsupported(sys.stdout)
-        stderr = self.none_if_fileno_unsupported(sys.stderr)
+        stdout = self.firstWithFilenoSupport(sys.stdout, sys.__stdout__)
+        stderr = self.firstWithFilenoSupport(sys.stderr, sys.__stderr__)
 
         self.solution_time = -clock()
         subprocess.check_call(proc, stdout=stdout, stderr=stderr)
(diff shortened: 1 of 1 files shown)


pulp/apis/scip_api.py
```diff
@@ -90,17 +97,21 @@ def actualSolve(self, lp):
...
 
-        self.solution_time = -clock()
-        subprocess.check_call(proc, stdout=sys.stdout, stderr=sys.stderr)
-        self.solution_time += clock()
+        stdout = self.firstWithFilenoSupport(sys.stdout, sys.__stdout__)
+        stderr = self.firstWithFilenoSupport(sys.stderr, sys.__stderr__)
+
+        self.solution_time = -resource.getrusage(resource.RUSAGE_CHILDREN).ru_utime
+        subprocess.check_call(proc, stdout=stdout, stderr=stderr)
+        self.solution_time += resource.getrusage(resource.RUSAGE_CHILDREN).ru_utime
```
pchtsp, line 114: I'm wondering why do you change the time measure? I'd ideally have the same way to measure time across solvers. How is it different from pulp's clock?

pchtsp (commented):
Thanks for the PR!
I'm guessing you've test it with SCIP and it works correctly. Sadly, we cannot test SCIP through github actions.
I just have a question regarding the clock you used.

pulp/apis/scip_api.py
wookenny: `clock()` measures only the time for the current process, excluding the child process  that is created with `check_call`.

For `PULP_CBC_CMD` it is also not working correctly, as I tested now: 
coin_time
and CBC logs
logs

Same for `COIN_CMD`

wookenny: Locally changed lines  https://github.com/coin-or/pulp/blob/a587e8e69d6629b3a55b2559a08179f019085253/pulp/pulp.py#L1739-L1741
and used `resource.getrusage(resource.RUSAGE_CHILDREN).ru_utime` instead of `clock()`.
For `GLPK_CMD`,  `PULP_CBC_CMD`, `COIN_CMD`, `SCIP_CMD' I get a correct `prob.solutionTime` and without this change, it is a wrong number close to 0.

wookenny: Currently:  `clock()`
clock
With ressource usage of child processes:
rusage

Interestingly, the `COIN_CMD` shows that the CPU time is measured, the logs of CBC show:
```
Result - Stopped on time limit

Objective value:                11.00000000
Lower bound:                    10.000
Gap:                            0.10
Enumerated nodes:               4414
Total iterations:               72247
Time (CPU seconds):             9.61
Time (Wallclock seconds):       6.50

Option for printingOptions changed from normal to all
Total time (CPU seconds):       9.66   (Wallclock seconds):       6.51
```

wookenny: I could create another PR changing this measuerement in `pulp.py` so that you could test it with other solver that you have access to.
I'm not entirely sure that every solver runs as a child process.

pchtsp:
Thanks again and sorry for the delayed response. Could we split this PR into one for the options /status for SCIP and another one with the clock time estimation at a "package level"? I never understood the way PuLP measured the solving time and I think that if your solution covers all cases (child process or not) then we should adopt it by putting it with the alias `clock` in the file `pulp.apis.core.py` file.

wookenny:
>  Could we split this PR into one for the options /status for SCIP and another one with the clock time estimation at a "package level"?
Absolutely reasonable. I'll close this PR and open two new ones, based on the current master.

PR #440, opened 2021-05-04
Handling user provided gaps in SCIP

wookenny:
Opened instead of #429,
this PR proposed to enable `absGap` and `relGap` for the scip api.

**Changes**
- passing arguments given to solver
- option handling aligned with the one for `coin_api.py`
- status for "gap limit reached" changed to OPTIMAL (optimal wrt. given gap)

**Example**
Bin packing MIP with known optimum (number of bins). 
https://github.com/wookenny/pulp/blob/6f36880525356466e49f0f97af61412a39764384/SCIP_handling_gap_reached.ipynb

### DIFF
--- a/pulp/apis/scip_api.py
+++ b/pulp/apis/scip_api.py
@@ -30,6 +30,7 @@ import os
 import io
 from .. import constants
 import sys
+import itertools
 
 
 class SCIP_CMD(LpSolver_CMD):
@@ -45,6 +46,8 @@ class SCIP_CMD(LpSolver_CMD):
         msg=True,
         options=None,
         timeLimit=None,
+        gapRel=None,
+        gapAbs=None,
         maxNodes=None,
     ):
         """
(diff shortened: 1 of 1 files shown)


pulp/apis/scip_api.py
```diff
@@ -141,6 +149,20 @@ def actualSolve(self, lp):
         self.delete_tmp_files(tmpLp, tmpSol)
         return status
 
+    def getOptions(self):
+        params_eq = dict(
+            timeLimit="set limits time {}",
```
pchtsp, line 154: one thing, could you take out the timeLimit here? in fact the timeLimit parameter is not stored in self.optionsDict. I know: it's anything but intuitive. Unfortunately, that's how it works with other CMD solvers too. If you check the `actualSolve` function, we assign the timeLimit from the `self.timeLimit` property. At least I want to keep all solvers "consistent".

pchtsp (changes requested):
Thanks again for the PR! There's just one line that should go away and I'll accept it.

PR #441, opened 2021-05-04
Measuring solutionTime including child processes

wookenny:
**Trigger:**
Variable `solutionTime` of `LpProblem` is not measuring the CPU time required by the solver correctly.

See the differences in `walltime` and `solutionTime` for the linked python [file](https://gist.github.com/wookenny/b9600a26a98b4fbbdd8329c00f55e7e7#file-report_optimization_time-py):

```
PULP_CBC_CMD
	objective:   8.00 (Optimal)
	time:   0.02s 	walltime:   6.43s
COIN_CMD
	objective:   8.00 (Optimal)
	time:   0.02s 	walltime:   6.72s
PULP_CHOCO_CMD
	objective:   4.00 (Optimal)
	time:   0.00s 	walltime:   0.41s
SCIP_CMD
	objective:  14.00 (Optimal)
	time:   0.11s 	walltime:   7.72s
```

**Changes:**
Using `resources` to get CPU time of child processes.

Resulting output:
```
PULP_CBC_CMD
	objective:   8.00 (Optimal)
	time:   6.22s 	walltime:   6.42s
COIN_CMD
	objective:   8.00 (Optimal)
	time:   6.42s 	walltime:   6.62s
PULP_CHOCO_CMD
	objective:   4.00 (Optimal)
	time:   0.83s 	walltime:   0.41s
SCIP_CMD
	objective:  14.00 (Optimal)
	time:   6.97s 	walltime:   7.26s
```

**Warning:** Only tested with the mentioned solver due to licenses. Please use the provided file and add available solver with appropriate number of bins.

### DIFF
--- a/pulp/apis/core.py
+++ b/pulp/apis/core.py
@@ -33,10 +33,14 @@ the current version
 import os
 import sys
 
-try:
-    from time import process_time as clock
-except ImportError:
-    from time import clock
+
+if os.name == "posix":
+    from ..utilities import resource_clock as clock
+else:
+    try:
+        from time import monotonic as clock
+    except ImportError:
+        from time import clock
 
 try:
     import configparser
@@ -49,7 +53,6 @@ except AttributeError:
 from .. import sparse
 from .. import constants as const
 
-import warnings
 import logging
 
 try:
@@ -91,7 +94,7 @@ class PulpSolverError(const.PulpError):
 
 # import configuration information
 def initialize(filename, operating_system="linux", arch="64"):
-    """ reads the configuration file to initialise the module"""
+    """reads the configuration file to initialise the module"""
     here = os.path.dirname(filename)
     config = Parser({"here": here, "os": operating_system, "arch": arch})
     config.read(filename)
@@ -239,7 +242,7 @@ class LpSolver:
     def actualResolve(self, lp, **kwargs):
         """
         uses existing problem information and solves the problem
-        If it is not implelemented in the solver
+        If it is not implemented in the solver
         just solve again
         """
         self.actualSolve(lp, **kwargs)
--- a/pulp/tests/test_pulp.py
+++ b/pulp/tests/test_pulp.py
@@ -5,6 +5,7 @@ from pulp.constants import PulpError
 from pulp.apis import *
 from pulp import LpVariable, LpProblem, lpSum, LpConstraintVar, LpFractionConstraint
 from pulp import constants as const
+from pulp.tests.bin_packing_problem import create_bin_packing_problem
 from pulp.utilities import makeDict
 import unittest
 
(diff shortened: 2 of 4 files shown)


wookenny:
Simple script for MIPs with adjustable complexity: https://gist.github.com/wookenny/b9600a26a98b4fbbdd8329c00f55e7e7#file-report_optimization_time-py

pulp/utilities.py
```diff
@@ -1,8 +1,12 @@
 # Utility functions
 import itertools
 import collections
+import resource
 
 
+def resource_clock():
```
wookenny, line 7: added to this file instead of `pulp.apis.core` to prevent circular imports

pchtsp:
Thanks for the PR. It would great to have a test on this. But I guess it's really hard to rest a certain duration since it depends on the machine, and sometimes the duration it's just random. Maybe we could test that "the duration of solving a really hard
/ large problem with a time limit is something around the time limit with some tolerance". We could even use your gist with a parametrized problem to play with "big enough" problems. What do you think? 
Also: can you pull the master branch? it has a new commit that runs the tests on PRs too.

wookenny:
Rebased on current's master.
I'll come up with a test as you proposed to test the time measurement.

As you said, better machines, more cores, improved solvers could all come up with a solution faster.
But it might work to create solver specific problem sizes so that they are done with the root relaxation but haven't found an optimal solution before the timelimit.
Since we measuere CPU time, it might also differ per solver what assertion to  formulate.
`timeLimit` could be interpreted by solver as wall or CPU time.

pulp/apis/core.py
```diff
@@ -33,10 +33,14 @@
 import os
 import sys
 
-try:
-    from time import process_time as clock
-except ImportError:
-    from time import clock
+
+if os.name == "posix":
+    from ..utilities import resource_clock as clock
```
wookenny, line 38: Under Linux, this is now measuring CPU time. 
If wall time is more appropriate to be silimar with windows, this case could be dropped entirely.

pulp/tests/test_pulp.py
```diff
@@ -1103,16 +1104,42 @@ def test_infeasible_problem__is_not_valid(self):
...
+        bins = solver_settings.get(self.solver.name)
+        if bins is not None:
+            prob = create_bin_packing_problem(bins=bins)
+            self.solver.timeLimit = time_limit
+            prob.solve(self.solver)
+
+            self.assertAlmostEqual(
+                prob.solutionTime,
+                time_limit,
+                delta=2,
```
wookenny, line 1138: Unfortunately, CBC shows higher variations in the measured time, so smaller `delta`s failed.
Could be tested with solver specific settings.

wookenny:
The current version passes the added test.
I'm totally open for change requests due to the added complexity. Maybe you prefer to simply measure walltime in all cases.

pchtsp:
thanks again for the PR! I *might* have preferred having Linux and Windows with the same functions, if possible. Would it be possible to just use `monotonic` for both? But the fact is I'm already ashamed of having taken too long to answer and approve this PR and it's already sufficiently good to add.

Issue #423, opened 2021-04-02
Repo does not follow flake8/isort/black standards


rimaddo:
Describe the new feature
--------------------------
These days formatting repos with flake8/isort (for python 3.5 and earlier) and black (for 3.6 and later) is standard for python repos.

However the pulp codebase does not match with these standards.

I would be interested in making this change, and could add a pre-commit-hook or similar so that this could be maintained in the future, but I wanted to understand if this was of interest to the maintainers / community beforehand, since I understand that it is quite a large change.

It would also be good to consider doing it in several steps perhaps, e.g. flake8 then isort then black perhaps.

pchtsp:
Hello, thanks for the suggestion.

Although I do think it's a good idea, I'm not 100% aware of what the changes would imply to be honest. Is it the variable naming convention? Or only spaces, commas, etc.?

The ideal would be pass the project through an automated tool (if available) and show the changes via a PR. And, as you say, eventually maybe have a hook or something.

If it's not too much of a hassle (i.e., automatic tool) to come up with a PR, I'll give it a look.

Thanks again.

rimaddo:
Okay thanks, I started with making a black pr as commented just above. This is the easiest change to make and also makes a big bulk of the changes. Then flake8 and isort could be made afterwards depending on thoughts.

Most of the changes from black in it as mentioned are around spaces, commas, quotes, line breaks etc.

Isort would impose rules on your import ordering / formatting etc.

Flake8 more generally on python standards eg should not import from `*`, use bare exceptions, etc.

You mentioned variable naming, that is something that is breaking pep8, since there are lots of variables and functions which are camel case as opposed to snake case. However, flake8 doesn't correct that as standard, you need to use something like `pep8-naming` to get linting corrections for this. I can understand though that this might be the most undesirable change to make, since the other linting changes described above will only impact your codebase, but renaming functions and variables to comply with pep8 naming will impact how people interact with your package as users will reference the current camelcase format.

pchtsp:
I'll close this issue as I think it was solved by the PR you did. Feel free to re-open it or open a new one if that's not the case!

PR #446, opened 2021-05-23
Add black to travis

rimaddo:
I previously ran black over the codebase here https://github.com/coin-or/pulp/pull/430, as a pr which came from issue https://github.com/coin-or/pulp/issues/423.

To up-keep this in a more automated manner I have also now added running it to travis, since your travis only has python 3.6+ versions (black cannot be installed with 3.5 or earlier).

This would take out the need for anyone to check that it is kept up to date manually, since the changes to keep it in line with the standard would always be displayed as part of the travis check.

I also made any changes to make it valid with travis since recent prs.

Many thanks!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,7 +36,7 @@ before_install:
 - sudo apt-get install -qq glpk-utils
 - if [[ $TRAVIS_PYTHON_VERSION ne pypy3 ]]; then
 - echo "Using black to check Python code"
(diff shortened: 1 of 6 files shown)


stumitchell:
Would probably be better if this was done with github actions rather than travis,

Travis is our legacy CI at the moment

rimaddo:
> Would probably be better if this was done with github actions rather than travis,
> 
> Travis is our legacy CI at the moment

I see sorry I didn't realise, will update to github actions!

rimaddo:
Changed to github actions in https://github.com/coin-or/pulp/pull/450

PR #450, opened 2021-06-06
Add black to GitHub actions

rimaddo:
I previously ran black over the codebase here https://github.com/coin-or/pulp/pull/430, as a pr which came from issue https://github.com/coin-or/pulp/issues/423.

To up-keep this in a more automated manner I have also now added running it to github actions.

This would take out the need for anyone to check that it is kept up to date manually, since the changes to keep it in line with the standard would always be displayed as part of the travis check.

I also made any changes to make it valid with black since recent prs.

Many thanks!

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -30,9 +30,9 @@ jobs:
     - name: Test with pulptest
       run: pulptest
     - name: Code Quality
-      if: ${{ matrix.python-version }} != 2.7
+      if: ${{ matrix.python-version >= 3.6 }}
       run: |
         echo Python version ${{ matrix.python-version }}
-        echo Version is not 2.7 ${{ matrix.python-version }} != 2.7
+        echo Version is not 2.7 ${{ matrix.python-version >= 3.6 }}
         pip install black
         black . --check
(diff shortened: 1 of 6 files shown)


pchtsp (commented):
Thanks again for the help. If I understand the current changes, they would check if the black linter passes or not. right? Then we would have to apply it manually and push again.
Is it possible to add a hook that automatically applies black changes to each push / PR ?

rimaddo:
> Thanks again for the help. If I understand the current changes, they would check if the black linter passes or not. right? Then we would have to apply it manually and push again.
> Is it possible to add a hook that automatically applies black changes to each push / PR ?

No problem, your correct this step only checks and if it is incorrect the changed aren't made and pushed by github actions. There are a couple of options for this,

1. in my last pr I did add [a pre-commit-hook that you can install](https://github.com/coin-or/pulp/blob/master/.pre-commit-config.yaml) if you're developing with python 3.6 or greater and is applied to your local commits. Perhaps this needs to be added to the documentation?
2. I've not done this before but in blacks own code base they have an [example github action where this is done automatically](https://github.com/psf/black/actions/runs/17913292/workflow), which is more what you're implying? This would mean that you would not have to apply manually and push again.

pchtsp:
Sorry for delaying this this much. I'm approving this PR. We can later decide if we want to do something on top of this.
Thanks!

rimaddo:
> Sorry for delaying this this much. I'm approving this PR. We can later decide if we want to do something on top of this.
> Thanks!

No worries, I did have an experimental branch trying to get the auto-commit working, but it was not as straight forward as the example. I will create a new pr if I manage to get it to work! Thanks

PR #468, opened 2021-08-09
Measuring solution time as walltime

wookenny:
## Changes
- Reporting `solutionTime` as wall time. 
- Additionally, measuring `solutionCpuTime` as done before.
- Warning when 

TODO: 
- [ ] Adjust tests and verify with additional solver.

### Example behavior:

This branch: https://gist.github.com/wookenny/d326022fedefd9266c101a2ef6365c9f
Master: https://gist.github.com/wookenny/924a2f61496547a975bd8f9a80dbdb7a

### DIFF
--- a/pulp/apis/coin_api.py
+++ b/pulp/apis/coin_api.py
@@ -164,6 +164,8 @@ class COIN_CMD(LpSolver_CMD):
             self.writesol(tmpMst, lp, vs, variablesNames, constraintsNames)
             cmds += "mips {} ".format(tmpMst)
         if self.timeLimit is not None:
+            if self.optionsDict.get('threads', 1) > 1:
+                warnings.warn("Beware: CBC uses timeLimit as cpu_time, not wall_time")
             cmds += "sec %s " % self.timeLimit
         options = self.options + self.getOptions()
         for option in options:
--- a/pulp/apis/coin_api.py
+++ b/pulp/apis/coin_api.py
@@ -164,7 +164,8 @@ class COIN_CMD(LpSolver_CMD):
             self.writesol(tmpMst, lp, vs, variablesNames, constraintsNames)
             cmds += "mips {} ".format(tmpMst)
         if self.timeLimit is not None:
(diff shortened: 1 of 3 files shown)


pchtsp (commented):
Thanks a lot for the help. And no need to apologize: we're all here working for free :  )

So if I understand, `time` measures wall_time and `clock` measures cpu_time. Now that I see all this I'm not sure I understand the use of clock in the rest of the pulp repository:

like for example SCIP: 
https://github.com/coin-or/pulp/blob/c769d683bb5f3212df9ea6ae2929e76b0efd9e4b/pulp/apis/scip_api.py#L133-L136
or gurobi:
https://github.com/coin-or/pulp/blob/c769d683bb5f3212df9ea6ae2929e76b0efd9e4b/pulp/apis/gurobi_api.py#L177-L180

Among other. They all import clock from `core.py` which would imply they measure cpu_time. But in theory since we measure it in `pulp.py` we should not need to measure it again in each solver...
Do you understand why there's this duplication?
If I can get away with it, I would take all references inside each solver and only leave those in `pulp.py`. But maybe I'm missing something.

wookenny:
> Among other. They all import clock from `core.py` which would imply they measure cpu_time. But in theory since we measure it in `pulp.py` we should not need to measure it again in each solver...
> Do you understand why there's this duplication?
> If I can get away with it, I would take all references inside each solver and only leave those in `pulp.py`. But maybe I'm missing something.

I haven't looked too closely at it. My initial ideas was that each solver might measure this differently, e.g., to accommodate that CBC can be either invoked as `subprocess` or used as a lib. Another idea would that this treatment allows to only measure the time required to solve the problem, excluding writing the LP, parsing parameter and parsing the solution after solving.

But even the variable names are not matching, `solveTime` and `solution_time `. 
I would try to remove them entirely.

pchtsp:
Don't worry about it. I'll try to unify the property names across solvers
and leave it like that. As you said, maybe someone wants to measure the
time to solve without the building.

Thanks again for your help!

Issue #451, opened 2021-06-11
SCIP interface does not read solution when problem not solved


tomaszG247:
Details for the issue
--------------------

#### What did you do?
Solving a problem with SCIP solver with time limit.

#### What did you expect to see?
After a time-out, a solution in accordance with SCIP logs

#### What did you see instead?
a zero solution for all variables that is the one computed by SCIP (I also checked the temporary file created by SCIP.

Useful extra information
-------------------------

Looking at the code, it seems to me the issue is at
https://github.com/coin-or/pulp/blob/master/pulp/apis/scip_api.py#L184-L185
these lines completely skip the solution read when the problem was not solved.
I would be tempeted to just remove constants.LpStatusNotSolved from the noSolutionStatus in 
https://github.com/coin-or/pulp/blob/master/pulp/apis/scip_api.py#L94, but I am not sure if this works in all cases.

The info below often helps, please fill it out if you're able to. :)

#### What operating system are you using?

- [x] Windows: ( _version:_10 )
- [ ] Linux: ( _distro:_ ___ )
- [ ] Mac OS: ( _version:_ ___ )
- [ ] Other: ___

#### I'm using python version:

- [ ] 2.7
- [ ] 3.4
- [ ] 3.5
- [x] 3.6
- [ ] Other: ___

#### I installed PuLP via:

- [x] pypi (python -m pip install pulp)
- [ ] github (python -m pip install -U git+https://github.com/coin-or/pulp)
- [ ] Other: ___ (conda?)

#### Did you also

- [ ] Tried out the latest github version: https://github.com/coin-or/pulp
- [x] Searched for an existing similar issue: https://github.com/coin-or/pulp/issues?utf8=%E2%9C%93&q=is%3Aissue%20

pchtsp:
Sorry but I'm at a miss. From what I understand, the `NO_SOLUTION_STATUSES` mean that a solution was not produced by this solver. As you can see the three options (infeasible, unbounded, not solve) really do imply that a solution was not found when the time limit was reached. So I don't see how do you expect to have a solution in these cases. Can you elaborate please?

tomaszG247:
Hello, thank for the reply. So what is happening in my case is that SCIP times out, meaning that he could not certify that the current solution he has is optimal within the gap I want. But he does have a solution. The solution is written out to a temporary file, but is not read by pulp back. 
[...]

PR #473, opened 2021-08-18
Correcting the solution status for SCIP in order to read integer feasible solutions

tomaszG247:
to address issue #451

### DIFF
--- a/pulp/pulp.py
+++ b/pulp/pulp.py
@@ -59,7 +59,7 @@ Choose a solver and solve the problem. ex:
 
 Display the status of the solution
 >>> const.LpStatus[status]
-'Optimal'
+'Optimal Solution Found'
 
 You can get the value of the variables using value(). ex:
 >>> value(x)
--- a/pulp/pulp.py
+++ b/pulp/pulp.py
@@ -311,23 +311,43 @@ class LpVariable(LpElement):
         self.expression = e
         self.addVariableToConstraints(e)
 
+    @classmethod
     def matrix(
-        self,
+        cls,
         name,
-        indexs,
+        indices=None,  # required param. enforced within function for backwards compatibility
         lowBound=None,
         upBound=None,
         cat=const.LpContinuous,
         indexStart=[],
+        indexs=None,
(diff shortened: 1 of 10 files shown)


pulp/pulp.py
```diff
@@ -59,7 +59,7 @@
 
 Display the status of the solution
 >>> const.LpStatus[status]
-'Optimal'
+'Optimal Solution Found'
```
pchtsp, line 62: I think this line shouldn't change.

pchtsp (commented):
Thanks for the PR! Could you please add a test that failed with the current version and passes with these changes? Feel free to add it for all solvers or only for SCIP in this file: https://github.com/coin-or/pulp/blob/master/pulp/tests/test_pulp.py

if you need to produce feasible but not-necessarily optimal solutions, feel free to use the `create_bin_packing_problem` function together with a timeLimit as in this test: https://github.com/coin-or/pulp/blob/469ef50a037075276e792a3bd5ae43db4fc4fc47/pulp/tests/test_pulp.py#L1148

tomaszG247:
. I added a test as you suggested. Actually it is not passing for both CBC and SCIP. It is clear for me that solution status should be 2.0 (Solution found), but I don't understand what is the expected behaviour for solver status. Is it 'not Solved' or 'Optimal' ? currnetly CBC is retruning optimal and SCIP Not Solved... both are somewhat inadequate IMO. Related to part of the discussion in #476. If you agree I can go with Optimal

pulp/pulp.py
tomaszG247: @pchtsp I corrected the pulp.py, must have made an auto-change

pchtsp (commented):
Yes, go with Optimal please. This way it will be consistent with other solvers.

Thanks

pchtsp:
Hello again! I've been thinking that it would be nice to add some unit tests that used some sample solution files from SCIP to test these changes.

Also, could you please try to sign the CLA again? I think it broke for a while.

Thanks : )

PR #481, opened 2021-09-15
Use classmethod as a decorator

ticosax:
Because it solves warning coming from linters and type checkers.


pchtsp (approved):

pchtsp:
Hello! Thanks for the PR. Could you please pass the "black" formatter in an additional commit? check here on how to to that: https://black.readthedocs.io/en/stable/

Issue #508, opened 2021-11-30
Crash when parsing CPLEX mipopt solutions


thiagojobson:
Details for the issue
--------------------

#### What did you do?

I tried to solve an integer model with CPLEX_CMD with a time limit set. I noticed that for some cases PuLP crashes when parsing feasible integer solutions.

A hidden feature that I just found out is that for some reason PuLP tries to optimize it twice. However, this is what happens when no integer solution is found in the first optimization, but one is found in the second optimization:

```
Read time = 0.01 sec. (2.62 ticks)
CPLEX> New value for time limit in seconds: 15
CPLEX> Version identifier: 20.1.0.0 | 2020-11-10 | 9bedb6d68
CPXPARAM_TimeLimit                               15
Tried aggregator 1 time.
MIP Presolve eliminated 74 rows and 1096 columns.
MIP Presolve modified 27553 coefficients.
Reduced MIP has 1982 rows, 29416 columns, and 138726 nonzeros.
Reduced MIP has 29385 binaries, 31 generals, 0 SOSs, and 0 indicators.
Presolve time = 0.13 sec. (108.17 ticks)
Tried aggregator 1 time.
Detecting symmetries...
Reduced MIP has 1982 rows, 29416 columns, and 138726 nonzeros.
Reduced MIP has 29385 binaries, 31 generals, 0 SOSs, and 0 indicators.
Presolve time = 0.09 sec. (76.76 ticks)
Probing fixed 12989 vars, tightened 58 bounds.
Probing changed sense of 1 constraints.
Probing time = 8.59 sec. (4415.97 ticks)
Clique table members: 7992.
MIP emphasis: balance optimality and feasibility.
MIP search method: dynamic search.
Parallel mode: deterministic, using up to 16 threads.
Root relaxation solution time = 0.97 sec. (576.62 ticks)

        Nodes                                         Cuts/
   Node  Left     Objective  IInf  Best Integer    Best Bound    ItCnt     Gap

      0     0       29.8256   206                     29.8256      126

Zero-half cuts applied:  2

Root node processing (before b&c):
  Real time             =   15.05 sec. (12647.53 ticks)
Parallel b&c, 16 threads:
  Real time             =    0.00 sec. (0.00 ticks)
  Sync time (average)   =    0.00 sec.
  Wait time (average)   =    0.00 sec.
                          ------------
Total (root+branch&cut) =   15.05 sec. (12647.53 ticks)


MIP - Time limit exceeded, no integer solution.
[...]

PR #512, opened 2021-12-10
Fix 508: does not crash when parsing mipopt solutions

thiagojobson:
The problem described in #508 is because solutions returned by CPLEX's mipopt solver lack the `dual` and `reducedCost` fields in the xml file.
I added try/catch so it does not crash anymore, but I'm not sure if, in this case, values should be set to `None` or `0`.


thiagojobson:
Hello 👋, any chance this gets reviewed?
The checks failed due to unrelated issues.

pchtsp:
Thanks for the contribution! And sorry for the delay response :  )

Could you please merge de master branch? the python version issues are already fixed there. 

Also, could you please add some unit test that failed previously and now passes with these changes? I think there's no present way to use a cplex license during unittests. But it could suffice to add a cplex test solution file to the repo and use it to run the tests.

pchtsp:
thanks again. I does look good. 
I just have one small caveat. The place where you put the test *is* the most intuitive. Unfortunately, since it "requires" having the solver CPLEX_CMD, it will not be run in github. One way of going around this would be to (1) change the `readsol` method to  be static and (2) move the test somewhere else (I'm not 100% sure where, though). What do you think?

thiagojobson:
Thanks @pchtsp !

I'm afraid that as I'm trying do add your suggestions I started to get the same problems in the workflow which seems to be unrelated to this PR.

thiagojobson:
Hello @pchtsp,

I've added a test case and looks like it is is running and passing.

I think there's no need to make `readsol` static [unless I'm missing something?] because we can always instantiate CPLEX_CM and normally call `readsol` even though CPLEX isn't present.

PR #514, opened 2021-12-16
Fix PyInstaller MacOS integration

### LINKED ISSUE #513: pulp fails to locate collected cbc executable on MacOS generated by PyInstaller
Details for the issue
--------------------

#### What did you do?
I tried to execute a simpleeval function that calls the pulp lib underneath it, that validates an expression

The expression being validated

An exception:
```error
'NoneType' object has no attribute 'actualSolve'
```

#### Useful extra information
I have a [bug in my project](https://github.com/GraphFilter/GraphFilter/issues/116) and at first the error was with  a corner case of PyInstaller. I was testing on Windows, and then when it was solved, I tried to replicate the solution on Mac. However, on MacOS I was having the same problem, and I thought it was an error with PyInstaller again, then I described the behaviour in [this issue](https://github.com/pyinstaller/pyinstaller/issues/6427). 

One insight a user told me in order to fix the bug is that, on [this line](https://github.com/coin-or/pulp/blob/0cf902579b8ca52cca09d66df5619095f8062f99/pulp/apis/core.py#L186) change:

```python
config_filename = os.path.join(DIRNAME, "..", PULPCFGFILE)
```
to

```python
config_filename = os.path.normpath(os.path.join(DIRNAME, "..", PULPCFGFILE)) 
```

because PuLP is failing to find the config file in spite of it being collected. That's because `DIRNAME = os.path.dirname(__file__)` points to a directory that does not exist in the frozen bundle (`sys._MEIPASS/pulp/apis)` because all of its content ends up in PYZ archive. But the config file, `sys._MEIPASS/pulp/apis/../pulp.cfg.osx = sys._MEIPASS/pulp/pulp.cfg.osx` does exist.

> As described [here](https://github.com/pyinstaller/pyinstaller/issues/6427#issuecomment-991961299)

-------------------------

- [ ] Linux: ( _distro:_ ___ )
- [x] Mac OS: ( version: 10.13.6 )
- [ ] Other: ___

- [ ] 2.7
- [ ] 3.4
- [ ] 3.5
- [ ] 3.6
- [x] Other: 3.9

#### I installed PuLP via:

- [x] pypi (python -m pip install pulp)
- [ ] Other: ___ (conda?)

#### Did you also

- [ ] Tried out the latest github version: https://github.com/coin-or/pulp
- [ ] Searched for an existing similar issue: https://github.com/coin-or/pulp/issues?utf8=%E2%9C%93&q=is%3Aissue%20

LBeghini:
Fixes #513

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -12,7 +12,11 @@ jobs:
         os: [ubuntu-latest, macOS-latest, windows-latest]
         exclude:
           - os: macOS-latest
-            python-version: [2.7, 3.5]
+            python-version: 2.7
+          - os: macOS-latest
+            python-version: 3.5
+          - os: macOS-latest
+            python-version: 3.6
 
     steps:
       - uses: actions/checkout@v1
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -11,7 +11,7 @@ jobs:
         python-version: [2.7, 3.5, 3.6, 3.7, 3.8]
         os: [ubuntu-latest, macOS-latest, windows-latest]
         exclude:
-          - os: macos-latest
+          - os: macOS-latest
             python-version: [2.7, 3.5]
 
     steps:
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -4,33 +4,35 @@ on: [push, pull_request]
 
 jobs:
   build:
-
     runs-on: ${{ matrix.os }}
     strategy:
       max-parallel: 21
       matrix:
         python-version: [2.7, 3.5, 3.6, 3.7, 3.8]
         os: [ubuntu-latest, macOS-latest, windows-latest]
+        exclude:
+          - os: macos-latest
+            python-version: [2.7, 3.5]
 
     steps:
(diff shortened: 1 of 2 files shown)


stumitchell:
Looks like the Mac-os testing has changed feel free to remove the version 2.7 and 3.5 testing from mac-os

LBeghini:
@stumitchell I don't know if I have permission to do that

stumitchell:
it is this file https://github.com/LBeghini/pulp/blob/master/.github/workflows/pythonpackage.yml

however I don't know how to remove specific combinations only as we probably still want python2.5 and 3.5 testing

pchtsp:
check here for an example on how to exclude:
https://github.com/pchtsp/pytups/blob/master/.github/workflows/pythonpackage.yml

LBeghini:
I removed the macOS configurations but apparently the action still ran the versions excluded. 🤔

stumitchell:
Maybe it is a poblem with the capitalisation? @LBeghini

pchtsp:
apparently exclude cannot take lists as input. Individual excludes need to be added. The current failures relate to black. Please run the black formatter on the files you edited as per the contribution guidelines: https://coin-or.github.io/pulp/develop/contribute.html#applying-the-black-linter-formatter

Issue #502, opened 2021-11-23
Allow capturing logs as string


jerr0328:
Describe the new feature
--------------------------
Similar to #188, I want to be able to capture the PuLP output and put it all in a single log line, rather than having it go to stdout directly. Instead of creating a temporary file and pointing PuLP to the log file, I would like to be able to get the log output as a string, similar to how you can do it with [subprocess.check_output()](https://docs.python.org/3/library/subprocess.html#subprocess.check_output). Alternatively, we could provide a file object (like StringIO) if that's easier.

Additional info
---------------
Please answer these questions before submitting your feature request.

#### Is your feature request related to an issue? Please include the issue number.
#188 

#### Does this feature exist in another product or project? Please provide a link.
Python has a native way of getting output from subprocess command calls: https://docs.python.org/3/library/subprocess.html#subprocess.check_output

pchtsp:
This make sense, many commercial solvers already provide a file object interface, if I remember correctly. I'd love to check a PR if you feel like helping out : )

rimaddo:
Is this feature implemented in https://github.com/coin-or/pulp/pull/525?

PR #525, opened 2022-03-02
Capture-solver-log

Xemnas0:
Asyncronously capture output of the solver and save it as a string.
Allows for both "msg" and "logPath" to be enabled.
Change applied only to class COIN_CMD

### DIFF
--- a/pulp/apis/coin_api.py
+++ b/pulp/apis/coin_api.py
@@ -187,7 +187,7 @@ class COIN_CMD(LpSolver_CMD):
         args.extend(cmds[1:].split())
 
         class SubProcessRunner:
-            def __init__(self, print_to_stdout: bool, logPath: str = None):
+            def __init__(self, print_to_stdout, logPath=None):
                 self.loop = self.get_event_loop()
                 self.output = []
                 self.full_output = None
--- a/pulp/apis/coin_api.py
+++ b/pulp/apis/coin_api.py
@@ -198,9 +198,9 @@ class COIN_CMD(LpSolver_CMD):
             def get_event_loop():
                 if sys.platform == "win32":
                     loop = asyncio.ProactorEventLoop()
-                    asyncio.set_event_loop(loop)
                 else:
-                    loop = asyncio.get_running_loop()
+                    loop = asyncio.new_event_loop()
+                asyncio.set_event_loop(loop)
                 return loop
 
             async def watch(self, stream):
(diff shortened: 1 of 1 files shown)


Xemnas0:
@pchtsp could you check this please?

pchtsp:
can you update with master again? the tests were failing for another reason. thanks.

Xemnas0:
Ah I see, python 2.7 does not support type annotations

pulp/apis/coin_api.py
```diff
@@ -176,31 +178,70 @@ def solve_CBC(self, lp, use_mps=True):
...
+                return_code = self.loop.run_until_complete(self.run_sub(args))
+                self.full_output = "\n".join(self.output)
+                self.loop.close()
+                return return_code
+
+        sub_process_runner = SubProcessRunner(print_to_stdout=self.msg, logPath=logPath)
+        return_code = sub_process_runner.run(args)
+
+        if return_code != 0:
+            print(return_code)
```
jerr0328, line 234: Should remove this, otherwise this can break standardized logging

pulp/apis/coin_api.py
```diff
@@ -176,31 +178,70 @@ def solve_CBC(self, lp, use_mps=True):
...
         log.debug(self.path + cmds)
         args = []
         args.append(self.path)
         args.extend(cmds[1:].split())
-        cbc = subprocess.Popen(args, stdout=pipe, stderr=pipe, stdin=devnull)
-        if cbc.wait() != 0:
-            if pipe:
-                pipe.close()
+
+        class SubProcessRunner:
```
jerr0328, line 189: I'm not the maintainer of the project so this is just an opinion, but to make testing easier, this class could be split out at a higher level and tested independently, rather than having the class declared inline.

pulp/apis/coin_api.py
```diff
@@ -176,31 +178,70 @@ def solve_CBC(self, lp, use_mps=True):
...
+
+        sub_process_runner = SubProcessRunner(print_to_stdout=self.msg, logPath=logPath)
+        return_code = sub_process_runner.run(args)
+
+        if return_code != 0:
+            print(return_code)
             raise PulpSolverError(
                 "Pulp: Error while trying to execute, use msg=True for more details"
                 + self.path
+                + "return_code={}\noutput={}".format(
```
jerr0328, line 238: Exception would be hard to read with return code having no space before it, maybe newline here? Or remove this? I'd also suggest rewording the full exception, because the suggestion to `use msg=True for more details` would print to stdout but if you capture the output via other means, it will mislead developers thinking that `msg=True` will give more details than already provided.

jerr0328 (commented):
Thanks, this indeed does seem to solve #502, as the solver output would be in the `solver_output` variable, though I haven't tried running with this branch. I've put some comments since I noticed some things that might cause problems for us if merged as-is.

PR #526, opened 2022-03-08
Fixed error when PATH is not set

snajder-r:
Default for a missing `PATH` should be an empty string, not an empty list. Otherwise trying to call `.split` on it will throw `AttributeError: 'list' object has no attribute 'split'`

### DIFF
--- a/pulp/apis/core.py
+++ b/pulp/apis/core.py
@@ -494,7 +494,7 @@ class LpSolver_CMD(LpSolver):
         if os.path.isabs(command):
             if os.path.exists(command) and os.access(command, os.X_OK):
                 return command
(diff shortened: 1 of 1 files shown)


pchtsp:
this makes a lot of sense, thanks. Can you please come up with a unittest to validate this and be sure it doesn't happen again? thanks again

pchtsp:
hello, can you pull the current master? I think the tests were failing for some other reason. Thanks and sorry for the delay.

pchtsp:
I'm closing since this issue was created, an alternative way to test the executable was found in 0c2b9418136406c7c54054f4ccbf0ec90b9787c4

PR #549, opened 2022-06-13
Xpress: flush after writing commands to stdin

### LINKED ISSUE #505: Issue using XPRESS solver
I have an issue using PULP with XPRESS.
XPRESS (v8.5) is correctly installed and works.
PULP correctly sees XPRESS as available solver.
When I try to solve a problem (ex sudoku) with XPRESS I see a log indicating the solver is initialized, but it never returns any solution.
It just seems to hang.
I have tried the "keepFiles" option without  additional luck.
Does it actually work ?

FYI : I am using PULP 2.5.0 with Python 3.8 on Windows 10 (x64)

octachrome:
Fixes #505. Xpress was hanging because it was waiting for input. The commands written to it by PuLP were being buffered. Flushing the stream after writing the commands fixes the problem.

### DIFF
--- a/pulp/apis/xpress_api.py
+++ b/pulp/apis/xpress_api.py
@@ -145,6 +145,7 @@ class XPRESS(LpSolver_CMD):
             xpress.stdin.write("GLOBAL\n")
         xpress.stdin.write("WRITEPRTSOL " + tmpSol + "\n")
         xpress.stdin.write("QUIT\n")
+        xpress.stdin.flush()
         if xpress.wait() != 0:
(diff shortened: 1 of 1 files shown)


pchtsp:
Thanks! I'm merging the changes. Is there any possibility of testing the xpress solver with small problems without getting a license? So we can add it to the pulp unit tests?

octachrome:
> Thanks! I'm merging the changes. Is there any possibility of testing the xpress solver with small problems without getting a license? So we can add it to the pulp unit tests?

You can download a free community edition of Xpress that you can use for testing: https://content.fico.com/xpress-optimization-community-license.

PR #554, opened 2022-06-27
checkLengthVars is not needed

berkorbay:
52 characters is very restricting and HiGHS can work pretty well with longer variable names. kwargs is included in the actualSolve method with a parameter of max_var_length (default: -1) to regulate character length if needed.


pchtsp:
We can directly remove the `lp.checkLengthVars(52)` line, instead of adding the parameter. I do not see the parameter being used in any other solver and I think it's not worth maintaining. What do you think?

berkorbay:
@pchtsp That would be my first attempt but I didn't want to break anything. I removed all the extra code and removed the checklengthvars line.

Issue #483, opened 2021-09-19
lpDot does not use the coefficients in LpAffineExpression


ytakashina:
Details for the issue
--------------------

#### What did you do?
```python
x = pd.Series(pl.LpVariable.dicts("x", range(5)), index=range(5))  # Series of LpVariable
y = 2 * x  # Series of LpAffineExpression
a = np.arange(5)
```

#### What did you expect to see?
```python
>>> pl.lpDot(y, a) 
LpAffineExpression([(x_1, 2), (x_2, 4), (x_3, 6), (x_4, 8)])
```

#### What did you see instead?

```python
>>> pl.lpDot(y, a)
LpAffineExpression([(x_1, 1), (x_2, 2), (x_3, 3), (x_4, 4)])
```

Useful extra information
-------------------------

The info below often helps, please fill it out if you're able to. :)

#### What operating system are you using?

- [x] Windows: ( _version:_ 10 )
- [ ] Linux: ( _distro:_ ___ )
- [ ] Mac OS: ( _version:_ ___ )
- [ ] Other: ___

#### I'm using python version:

- [ ] 2.7
- [ ] 3.4
- [ ] 3.5
- [ ] 3.6
- [x] Other: 3.7

#### I installed PuLP via:

- [x] pypi (python -m pip install pulp)
- [x] github (python -m pip install -U git+https://github.com/coin-or/pulp)
- [ ] Other: ___ (conda?)

#### Did you also

- [x] Tried out the latest github version: https://github.com/coin-or/pulp
- [x] Searched for an existing similar issue: https://github.com/coin-or/pulp/issues?utf8=%E2%9C%93&q=is%3Aissue%20

sdementen:
fyi, as a workaround you can use `y @ a`  instead of lpDot

frrad:
Should be fixed by https://github.com/coin-or/pulp/pull/769

ytakashina:
fixed, thx!

PR #572, opened 2022-08-22
pulp.lpDot update

nariaki3551:
to address issue #483

With this change, the following lpDot operation behaves as expected.

```python
import pandas as pd
import numpy as np
import pulp as pl

x = pd.Series(pl.LpVariable.dicts("x", range(3)), index=range(3))  # Series of LpVariable
y = 2 * x  # Series of LpAffineExpression [2*x_0, 2*x_1, 2*x_2]
a = np.arange(3) + 1  # [1, 2, 3]

print(pl.lpDot(y, a))
>>> 2*x_0 + 4*x_1 + 6*x_2   # but, current version is x_0 + 2*x_1 + 3*x_2
```


pchtsp:
could you add a unit test to show what did not work and how it works now? thanks.

nariaki3551:
Thanks for your comment!
I added a test for this change.
With this change, lpDot outputs the correct LpAffineExpression even when an LpAffineExpression is given to lpDot.

pchtsp:
thanks and sorry for the late reply. Could you take out PSCIP_CMD from this PR? so I can add only the lpDot use? thanks. You can add a new PR with the new solver.

nariaki3551:
I have completed the requested changes. The PSCIP_CMD addition has been removed from this pull request, focusing solely on the lpDot updating as you suggested.

I apologize for my delayed response.
Thanks.

frrad:
Can probably be closed.

PR #580, opened 2022-09-22
Add support for MaxNode option with CBC solver

E-Aho:
This commit adds support for the `maxNodes` option for the CBC solver. Includes a simple test to make sure the option is implemented correctly.

I originally tried to create a test to show that using the `maxNodes` option would cause the CBC solver to terminate upon reaching that number of nodes, but couldn't find an intuitive way to read the `Enumerated Nodes` or "stopping reason" outputs of the sovler in test cases.

### DIFF
--- a/pulp/apis/coin_api.py
+++ b/pulp/apis/coin_api.py
@@ -52,6 +52,7 @@ class COIN_CMD(LpSolver_CMD):
         maxSeconds=None,
         gapRel=None,
         gapAbs=None,
+        maxNodes=None,
         presolve=None,
         cuts=None,
         strong=None,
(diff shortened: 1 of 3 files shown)


E-Aho:
Pinging @pchtsp to check if there's any update on this PR :)

For reference, I've been using the change in this fork since raising the PR and it has been working fine

pulp/apis/core.py
```diff
@@ -205,6 +205,7 @@ def __init__(
         self.msg = msg
         self.options = options
         self.timeLimit = timeLimit
+        self.maxNodes = maxNodes
```
pchtsp, line 208: as the comment two lines below says, the maxNodes attribute is stored inside the `self.optionsDict` attribute.

pulp/apis/coin_api.py
```diff
@@ -167,6 +170,8 @@ def solve_CBC(self, lp, use_mps=True):
             cmds += f"mips {tmpMst} "
         if self.timeLimit is not None:
             cmds += f"sec {self.timeLimit} "
+        if self.maxNodes is not None:
```
pchtsp, line 173: As the other comment says, this attribute does not exist. You should thus do something like:
```
maxNodes = self.optionsDict.get("maxNodes")
if maxNodes is not None:
    cmds += f"maxNodes {maxNodes} "
```

pchtsp (changes requested):

pchtsp:
sorry for the troubles, I've added the maxNodes argument as explained in the comments https://github.com/coin-or/pulp/pull/690. I'm closing the PR.

PR #555, opened 2022-06-30
HiGHS_CMD update

samiit:
# Reference issue

[553](https://github.com/coin-or/pulp/issues/553)

# What does this implement?

An interface for using the [HiGHS solver](https://www.highs.dev/) for MILP problems, with a modification to an earlier [PR](https://github.com/coin-or/pulp/pull/477).

# Any other comments
Requirements:

We would need to get the necessary binaries for the HiGHS solver from [here](https://github.com/JuliaBinaryWrappers/HiGHS_jll.jl/releases). The Windows binaries e.g., are [here](https://github.com/JuliaBinaryWrappers/HiGHS_jll.jl/releases/download/HiGHS-v0.3.3%2B0/HiGHS.v0.3.3.x86_64-w64-mingw32-cxx11.tar.gz).
Usage:

We can now create a HiGHS_CMD object, just like for other solvers, like, Cbc
```
import pulp as pl
highs_solver = pl.HiGHS_CMD(path = "<file location of HiGHS solver binary>", timeLimit=60, msg=True)
prob = pl.LpProblem("Test")

...

prob.solve(highs_solver)
```

PR #593, opened 2022-10-27
Update HiGHS_CMD

chriswasser:
Hi everyone,

following up on [the discussion](https://github.com/coin-or/pulp/pull/555#issuecomment-1274470458) in a previous PR of the HiGHS_CMD solver in PuLP, I would like to contribute the following improvements to the HiGHS_CMD solver:
- Add support for the solver options `gapRel`, `gapAbs`, `threads`, and `logPath`
- Add support for (native) maximization in MPS files (before problems were transformed into equivalent min. problems)
- Allow config file options to be specified (besides the already supported CLI options) which enables extensibility for future HiGHS updates without requiring immediate updates to PuLP. The supported syntax for passing CLI and / or file options via the solvers `option` parameter is:
  - assume that all options require an argument: If the given option does not contain an equal sign `=`, we assume that the parameter is the next item in the option list. Thus, it is possible to specify either `options=["--presolve", "on"]` or `options=["--presolve=on"]`.
  - a leading dash `-` indicates CLI options (other options are interpreted as file options)
- Update the README to reflect that HiGHS is now supported

As previously indicated, I have not tried to remain backwards compatible with Python 2 and <3.7 since they have all reached there [end of life](https://devguide.python.org/versions/#versions) (some a really long time ago). Maybe this would be a good time to migrate to a more modern Python version range i.e. >=3.7!? This would allow PuLP to eliminate a few outdated code paths and to adopt modern features of the standard library as well as allow for the integration of type hints which improve code completion and overall developer friendliness. That’s primarily for the maintainers of PuLP to decide though.

Thanks for considering this PR and providing this awesome package 🎉

Greetings

Christian


pchtsp:
Thanks for the PR. Please, re-merge to run tests again.

chriswasser:
Rebased my branch on top of the current master now 👍

PR #594, opened 2022-10-27
Update SCIP_CMD, Add FSCIP_CMD & SCIP_PY

chriswasser:
Hi everyone,

I would like to contribute the following extensions for the SCIP API:
- Add `threads` and `logPath` options to SCIP_CMD
- Add a warning for the `mip` option as it was previously silently ignored by SCIP_CMD
- Add FSCIP_CMD as a parallelized SCIP_CMD variant
- Add SCIP_PY which uses PySCIPOpt (the Python API of SCIP)
- Allow config file options to be specified (besides the already supported CLI options) which enables extensibility for future SCIP / FSCIP updates without requiring immediate updates to PuLP. The supported syntax for passing CLI and / or SCIP / FSCIP file options via the solvers option parameter is:
  - a leading dash `-` indicates CLI options (other options are interpreted as file options)
  - all options - whether provided via the CLI or in a file - require an argument:
    - CLI options are passed to the shell command as a separate argument
    - SCIP file options contain a slash `/` and are delimited by an equal sign `=`
    - FSCIP file options do not contain a slash `/` and are also delimited by an equal sign `=`
    - If the given option does not contain an equal sign =, we assume that the parameter is the next item in the option list. Thus, it is possible to specify either `options=["lp/presolving", "TRUE"]` or `options=["lp/presolving=TRUE"]`
- Add tests for the new solvers
- Update the README to reflect that FSCIP is now supported

As with [my other PR](https://github.com/coin-or/pulp/pull/593), I have not tried to remain backwards compatible with Python 2 and <3.7 since they have all reached there [end of life](https://devguide.python.org/versions/#versions) (some a really long time ago). Maybe this would be a good time to migrate to a more modern Python version range i.e. >=3.7!? This would allow PuLP to eliminate a few outdated code paths and to adopt modern features of the standard library as well as allow for the integration of type hints which improve code completion and overall developer friendliness. That’s primarily for the maintainers of PuLP to decide though.

Thanks for considering this PR and providing this awesome package 🎉

Greetings

Christian

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -4,7 +4,7 @@ pulp
     :target: https://travis-ci.org/coin-or/pulp
 
 PuLP is an LP modeler written in Python. PuLP can generate MPS or LP files
-and call GLPK_, COIN-OR CLP/`CBC`_, CPLEX_, GUROBI_, MOSEK_, XPRESS_, CHOCO_, MIPCL_, SCIP_ to solve linear
+and call GLPK_, COIN-OR CLP/`CBC`_, CPLEX_, GUROBI_, MOSEK_, XPRESS_, CHOCO_, MIPCL_, SCIP_/FSCIP_ to solve linear
 problems.
 
 Installation
@@ -147,3 +147,4 @@ The home page for the documentation is ``doc/build/html/index.html`` which can b
 .. _CHOCO: https://choco-solver.org/
 .. _MIPCL: http://mipcl-cpp.appspot.com/
 .. _SCIP: https://www.scipopt.org/
+.. _FSCIP: https://ug.zib.de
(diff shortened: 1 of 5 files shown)


pchtsp:
Thanks for the PR.

Can you get the master branch and resolve conflicts? we can then re-run the tests since we now have taken out the py 2.7 tests from the master. Tests should pass after merging.

And sorry for the delay in response.

woutkonings:
Hi, I opened the following [issue](https://github.com/coin-or/pulp/issues/658), as for me FSCIP_CMD is not recognized in pulp version 2.7.0.

PR #595, opened 2022-10-27
Upgrade to Python 3.7+

### LINKED ISSUE #552: Drop support for unsupported Python versions (2.7 and 3.4-3.6)
I would want to open the discussion about dropping unsupported Python versions for the next release, and especially Python 2. By supporting only supported Python versions for new releases, new features can be adopted and no maintenance effort is poured into supporting old Python versions. Developers can contribute using modern Python syntax.

Practically, this would currently mean setting the minimum supported Python version to 3.7 for the next release, dropping support for Python 2.7, 3.4, 3.5 and 3.6.

For currently supported Python versions, see: https://endoflife.date/python

If consensus is reached about supporting Python 3.7+ for the next release:

- [pyupgrade](https://github.com/asottile/pyupgrade) can be used to update the codebase for the most part to modern Python standards
- In setup.py `python_requires=">=3.7"` should be added
- The Readme should be updated

EwoutH:
This PR upgrades the code base to Python 3.7, including:
 - Requiring Python 3.7 in setup.py
 - Upgrades to Python 3.7+ syntax using Pyupgrade and flynt
 - Updating the Readme
 - Updating the packaging CI
See each individual commit for details.

Python 3.6 support has been dropped almost a year ago and Python 2.7 is end-of-life for almost 3 years now.

These changes affect only future releases, older releases like 2.6.0 can still be used with the old Python versions.

Closes #552.

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -8,18 +8,13 @@ jobs:
     strategy:
       max-parallel: 21
       matrix:
-        python-version: [2.7, 3.7, 3.8, 3.9, '3.10']
+        python-version: ['3.7', '3.8', '3.9', '3.10']
         os: [ubuntu-latest, macOS-latest, windows-latest]
-        exclude:
-          - os: macOS-latest
-            python-version: 2.7
-          - os: windows-latest
-            python-version: 2.7
 
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v3
       - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v1
+        uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
       - name: install glpk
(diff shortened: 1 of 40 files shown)


stumitchell:
Yeah we have kept 2.7 support for a long time. But it is probably holding us back now. Can we confirm that python 2.7 users would still be able to find and install pulp through pip etc. i.e. it will point them to a previous version.
I think this should also be a Major Version number increment
@pchtsp your thoughts?

EwoutH:
> Can we confirm that python 2.7 users would still be able to find and install pulp through pip etc. i.e. it will point them to a previous version.

Yes. pip searches for the `python_requires` tag I added to setup.py. Previous releases didn’t have that, so pip will try to install 2.6.0 on all previous Python versions. But all new releases will only be installed on Python 3.7 and higher.

See https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#python-requires

pchtsp:
sorry for the delay, I've been quite busy. I agree we can drop python 2.7. I'm going to release a last version with the current master. We can then get all the other PRs that assume python >= 2.7, and we can change to version 3.0, while updating docs.

PR #611, opened 2022-12-11
CI: Add Python 3.11 to the job matrix

EwoutH:
Add Python 3.11 to the job matrix, and also allow [manually triggering](https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow) a CI run.


EwoutH:
The Python 3.11 jobs fail currently because [highspy](https://github.com/ERGO-Code/HiGHS) hasn't published Python 3.11 wheels [on PyPI](https://pypi.org/project/highspy/) yet. I requested those, so we can track that here: https://github.com/ERGO-Code/HiGHS/issues/1042

PR #614, opened 2022-12-27
Implementing TimeLimit for highspy

apschmidt:
Time limit was ignored in previous version for highspy, adding timeLimit functionality.

### DIFF
--- a/pulp/apis/highs_api.py
+++ b/pulp/apis/highs_api.py
@@ -259,8 +259,8 @@ class HiGHS(LpSolver):
             *args,
             **solverParams,
         ):
-            super().__init__(mip, msg, timeLimit, gapRel=gapRel, *args, **solverParams)
-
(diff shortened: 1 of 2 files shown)


apschmidt:
Seems to be an issue with ubuntu highspy implementation since Windows passes all tests

pchtsp:
Mmm strange. Maybe you can check with the highs solver developers? Also, can you sign the cla? Thanks a lot for the contribution.

MLopez-Ibanez:
> Seems to be an issue with ubuntu highspy implementation since Windows passes all tests

Looking at the logs, it seems highspy is not install nor tested for Windows so that would explain it.

pchtsp:
It doesn't seem to pass the tests because the `timeLimit` check fails in Ubuntu actually. Are you sure it works?
image

pchtsp:
I've merged the other PR #641 which addresses the issues with checking the timelimit in this one. Thanks for the work though!

PR #622, opened 2023-01-11
Change behaviour of original bounds for Binary variables

### LINKED ISSUE #620: Counterintuitive behaviour with Binary variable bounds
Details for the issue
--------------------

Counterintuitive behaviour of bounds with Binary variable. This is specially bad when checking if a variable is Binary (`self.isBinary()` method returns False when the variable has been fixed).

At least, I would expect that the private attributes `self._lowbound_original = 0` and `self._upbound_original = 1`

#### What did you do?
```
import pulp
X = pulp.LpVariable("X",cat="Binary")

print("Before fixing value")
print(f"{X.lowBound=}")
print(f"{X._lowbound_original=}")
print(f"{X.isConstant()=}")


X.setInitialValue(1)
X.fixValue()

print("After fixing value to 1")
print(f"{X.lowBound=}")
print(f"{X._lowbound_original=}")
print(f"{X.isConstant()=}")
print(f"{X._lowbound_unfix=}")
```

```
Before fixing value
X.lowBound=0
X._lowbound_original=0
X.isConstant()=False
After fixing value to 1
X.lowBound=1
X._lowbound_original=0
X.isConstant()=True
X._lowbound_unfix=0
```

```
Before fixing value
X.lowBound=0
X._lowbound_original=None
X.isConstant()=False
After fixing value to 1
X.lowBound=1
X._lowbound_original=None
X.isConstant()=True
X._lowbound_unfix=0
```

And equivalently with the upper bound.

Useful extra information
-------------------------

- [x] Mac OS: ( _version: 13.0.1 )

- [x] Other: 3.9.13

#### I installed PuLP via:

- [x] pypi (python -m pip install pulp)

#### Did you also

- [x] Tried out the latest github version: https://github.com/coin-or/pulp
- [x] Searched for an existing similar issue: https://github.com/coin-or/pulp/issues?utf8=%E2%9C%93&q=is%3Aissue%20

davidggphy:
In Binary variables, if the inputs `lowBound` and `upBound` are the default None values, in the `__init__` the `self.lowBound` and `self.upBound` are set to 0,1. But the `self._lowbound_original`, `self._upbound_original` are set to None.  In the rest of the variables (Continuous, Integer) these two pairs of attributes coincide.

This is, the behaviour is equivalent to
```
self.lowBound = 0
self.upBound = 1
self._lowbound_original = None
self._upbound_original = None
```

It seems more consistent to have
```
self.lowBound = self._lowbound_original = 0
self.upBound = self._upbound_original  = 1
```

closes #620

### DIFF
--- a/pulp/pulp.py
+++ b/pulp/pulp.py
@@ -264,8 +264,8 @@ class LpVariable(LpElement):
         self.varValue = None
         self.dj = None
         if cat == const.LpBinary:
-            self.lowBound = 0
-            self.upBound = 1
+            self._lowbound_original = self.lowBound = 0
+            self._upbound_original = self.upBound = 1
             self.cat = const.LpInteger
         # Code to add a variable to constraints for column based
         # modelling.


davidggphy:
Looks like this breaks other things

PR #641, opened 2023-04-01
HiGHS API interface improvements including time_limit

aphi:
For the HiGHS API interface:

1. **Add the setting of a time_limit and significantly speed up solution processing**
  Picking up from the stale branch and PR (https://github.com/coin-or/pulp/pull/614), where the implementation is correct but the tests failed. The reason why took some digging...
  
    The time limit test for all the solvers is actually on the end-to-end time including extraction of solution values.

    In this case the solver terminated correctly after 10 seconds, but ~163s were spent in extracting the solution on my machine (or 320s in CI). Specifically, calling the __getitem__ method (i.e. []) in `solution.col_value[var.index]` is a very slow operation (presumably some type conversions via pybind11).

    For the 134k variables of the bin packing problem, **this took 163s and failed the test, but could be reduced very easily to 0.004s** 🎉 just by copying solution.col_value to a Python list first, for which index lookups are of course fast.

    Demo script in a gist if curious: https://gist.github.com/aphi/4b2f7cfc7ccc52465bead3c24ee9c3f4

    image

2. **Enable setting of any other paramters and clean up init signature**
  Clean up: `epgap` was unused and deprecated in favour of `gapRel`. `gapRel` as its own argument was also unsued and misleading. It was passed as an *args to the base class, which ignored the *args.

    `gapRel` when supplied via the **kwargs to optionsDict functions the same. But now the user can also specify any parameter for HiGHS, just the same handling as the Gurobi API.

    I've tested this and verified it works with other parameters like `threads` and `parallel` from the [HiGHS options](https://github.com/ERGO-Code/HiGHS/blob/master/src/lp_data/HighsOptions.h)
    
 3. Some minor tidyups

Happy to make any changes - just let me know.

P.S. I contributed to this project back before it was on GitHub (2007~2011) under the guidance of stumitchell. So great to see its immense popularity today!

### DIFF
--- a/pulp/apis/highs_api.py
+++ b/pulp/apis/highs_api.py
@@ -254,14 +254,11 @@ class HiGHS(LpSolver):
             mip=True,
             msg=True,
             timeLimit=None,
-            epgap=None,
-            gapRel=None,
             warmStart=False,
             logPath=None,
-            *args,
             **solverParams,
         ):
-            super().__init__(mip, msg, timeLimit=timeLimit, gapRel=gapRel, *args, **solverParams)
+            super().__init__(mip, msg, timeLimit=timeLimit, **solverParams)
 
         def available(self):
             return True
@@ -359,7 +356,7 @@ class HiGHS(LpSolver):
 
             for var in lp.variables():
                 var.modified = False
-                
+
             for constraint in lp.constraints.values():
                 constraint.modifier = False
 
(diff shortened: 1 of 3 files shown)


pchtsp:
thanks! I will close the other PR then

stumitchell:
@aphi hi long time no see :-) 

there seems to be a failing test in highs_api.py in the current master branch are you able to look into it, could be a background dependency issue, as the error showed up in an unrelated commit.

https://github.com/coin-or/pulp/actions/runs/5561352717/job/15064721500

Thanks

PR #656, opened 2023-05-31
Parsing objective sense in MPS file

### LINKED ISSUE #655: reading MPS file doesn't parse the objective sense
Describe the new feature
--------------------------
When reading an MPS file, the objective sense is not parsed. The objective sense has to been passed to `fromMPS` as a parameter before reading the instance.

I would like add to the `readMPS` method to parse the objective sense, if it is specified in the file. If the objective sense is not specified, then the sense defaults to `LpMinimize`.

stephenjmaher:
Fixes #655

This PR makes it possible to set the objective sense in an MPS file. Currently, the objective sense in the MPS file is ignored. The sense needs to be supplied to `fromMPS` as a parameter.

This change does the following:
- the `sense` parameter in `LpProblem.fromMPS` now defaults to `None`.
- the `readMPS` methods checks for the header `OBJSENSE` in the MPS file. If the header is encounter there are two options:
```
OBJSENSE   MAX
``` 
or
```
OBJSENSE
   MAX
```
both options are handled in the `readMPS` method.
- If no objective sense is specified in the MPS file, then sense defaults to `LpMinimize`.
- If an objective sense is passed to `LpProblem.fromMPS`, then this overrides the sense that is specified in the MPS file.

### DIFF
--- a/pulp/mps_lp.py
+++ b/pulp/mps_lp.py
@@ -152,7 +152,7 @@ def readMPS(path, sense, dropConsNames=False):
                 if line[1] not in bnd_names:
                     bnd_names.append(line[1])
             elif mode == CORE_FILE_OBJSENSE_MODE:
-                parameters['sense'] = const.LpMaximize if line[1] == 'MAX'\
+                parameters['sense'] = const.LpMaximize if line[0] == 'MAX'\
                         else const.LpMinimize
     constraints = list(constraints.values())
     if dropConsNames:
--- a/pulp/mps_lp.py
+++ b/pulp/mps_lp.py
@@ -7,6 +7,7 @@
 import re
 from . import constants as const
 
+CORE_FILE_OBJSENSE_MODE = "OBJSENSE"
 CORE_FILE_ROW_MODE = "ROWS"
 CORE_FILE_COL_MODE = "COLUMNS"
 CORE_FILE_RHS_MODE = "RHS"
@@ -34,7 +35,9 @@ def readMPS(path, sense, dropConsNames=False):
     This dictionary can be used to generate an LpProblem
 
     :param path: path of mps file
-    :param sense: 1 for minimize, -1 for maximize
+    :param sense: 1 for minimize, -1 for maximize, can be None. If None and the
+    MPS file doesn't set the sense, then the objective sense defaults to
+    minimize.
     :param dropConsNames: if True, do not store the names of constraints
     :return: a dictionary with all the problem data
     """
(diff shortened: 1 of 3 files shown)


pulp/mps_lp.py
```diff
@@ -83,6 +86,12 @@ def readMPS(path, sense, dropConsNames=False):
                     mode = CORE_FILE_BOUNDS_MODE_NAME_GIVEN
                 else:
                     mode = CORE_FILE_BOUNDS_MODE_NO_NAME
+            elif parameters['sense'] is None and line[0] == CORE_FILE_OBJSENSE_MODE:
+                if len(line) > 1:
+                    parameters['sense'] = const.LpMaximize if line[1] == 'MAX'\
+                            else const.LpMinimize
```
pchtsp, line 92: Aren't you missing a `mode=CORE_FILE_OBJSENSE_MODE` here too?

pulp/mps_lp.py
```diff
@@ -83,6 +86,12 @@ def readMPS(path, sense, dropConsNames=False):
...
+                else:
+                    mode = CORE_FILE_OBJSENSE_MODE
```
pchtsp, line 94: I'm not sure why would we have the same mode regardless of the if clause. Maybe we should put the assignment before the `len(line)>1`  and take out the `else` clause? Or maybe I don't understand the idea.

pulp/mps_lp.py
```diff
@@ -142,12 +151,20 @@ def readMPS(path, sense, dropConsNames=False):
                 readMPSSetBounds(line, variable_info)
                 if line[1] not in bnd_names:
                     bnd_names.append(line[1])
+            elif mode == CORE_FILE_OBJSENSE_MODE:
+                parameters['sense'] = const.LpMaximize if line[0] == 'MAX'\
+                        else const.LpMinimize
```
pchtsp, line 156: shouldn't we test if `parameters['sense']` is None? or if we provided an argument to the function (i.e., via the `sense` argument?

pchtsp (changes requested):
Thanks for the update. I'm still unsure if we should where we should check for None and where to assign the mode of the line. I've left some comments.

pulp/mps_lp.py
stephenjmaher: Not in this case. This line is catching the case where you have the objective sense provided in the same line as the keyword `OBJSENSE`, such as
```
OBJSENSE MAX
```
In this case we don't need to enter the OBJSENSE mode.

pulp/mps_lp.py
stephenjmaher: Following on from the previous comment, we have to enter the `CORE_FILE_OBJSENSE_MODE` is the objective sense is not given in the same line as the `OBJSENSE` keyword. This is catching the case
```
OBJSENSE
  MAX
```
The check of `len(line) > 1` differentiates between these two cases.

pulp/mps_lp.py
stephenjmaher: We could put in a check here for whether `parameters['sense']` is `None`. However, in the current code the check is not necessary since it is only possible to enter the `CORE_FILE_OBJSENSE_MODE` mode if `parameters['sense']` is `None`.

We could put in an assert or throw an exception to protect against future code changes? Something like (after line 154)
```
if parameters['sense'] is not None
    raise Exception("The supplied objective sense will be overwritten by the MPS file objective sense")
```

pchtsp:

PR #659, opened 2023-06-14
Add COPT support for PuLP

wujianjack:
This PR adds COPT support for PuLP, including `COPT_CMD()`, `COPT_DLL()` and `COPT()` solvers. 

The `COPT_CMD()` and `COPT_DLL()` solvers depend on `copt_cmd` and COPT shared library respectively, so a full installation of COPT package is required. You can obtain the latest packages from https://github.com/COPT-Public/COPT-Release.

The `COPT()`  solver depends on `coptpy`, you can install with `pip install coptpy`.

For larger instances, a valid COPT license is required, you can apply one from https://www.shanshu.ai/copt.

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -30,6 +30,10 @@ jobs:
         if: matrix.os == 'ubuntu-latest' || matrix.os == 'macOS-latest'
         run: |
           pip install highspy
+      - name: Install coptpy
+        if: matrix.os == 'ubuntu-latest' || matrix.os == 'macOS-latest' || matrix.os == 'windows-latest'
+        run: |
+          pip install coptpy
       - name: Test with pulptest
         run: pulptest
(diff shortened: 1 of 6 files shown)


pchtsp:
this is great! It would be great to add these solvers to the tests, like here:

https://github.com/coin-or/pulp/blob/a7684ac90d93dc3bca382d90841a9ab5f398b50f/pulp/tests/test_pulp.py#L1479-L1481

And also it would be even greater to add the installation of the python package in the integrated tests:

https://github.com/coin-or/pulp/blob/a7684ac90d93dc3bca382d90841a9ab5f398b50f/.github/workflows/pythonpackage.yml#L29-L32

by adding another step with the installation of the package.

pchtsp:
hello! it seems that the tests are falling. Could you take a look? Are you sure the `COPT_HOME` env variable is set?

PR #660, opened 2023-06-14
Add setting randomseed support for CPLEX_PY

MBradbury:
The [random seed](https://www.ibm.com/docs/en/icos/12.8.0.0?topic=parameters-random-seed) of a CPLEX solver can be set by updating the `parameters.randomseed` variable. This PR adds support for users to set the seed when the CPLEX_PY solver is created.

### DIFF
--- a/pulp/apis/cplex_api.py
+++ b/pulp/apis/cplex_api.py
@@ -303,6 +303,7 @@ class CPLEX_PY(LpSolver):
             logPath=None,
             epgap=None,
             logfilename=None,
+            randomseed=None,
         ):
             """
             :param bool mip: if False, assume LP even if integer variables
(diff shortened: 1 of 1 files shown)


pchtsp:
Is there a particular reason to do it through the initialization of the solver instead of the current supported way? Ideally, we want to keep an interface as simple as possible and as uniform across solvers as possible. Currently we do not pass seeds for other solvers.

MBradbury:
What is the existing way? As far as I could see there was no way to pass a seed to CPLEX. If there is an existing API that would be great.

pchtsp:
Check https://coin-or.github.io/pulp/guides/how_to_configure_solvers.html#using-solver-specific-functionality 

```
import pulp

x = pulp.LpVariable('x', lowBound=0)
prob = pulp.LpProblem('name', pulp.LpMinimize)
prob += x

solver = pulp.CPLEX_PY()
solver.buildSolverModel(prob)
# you can now edit the object or do something with it before solving
# for example, load a MIP_START file for CPLEX_PY:
solver.solverModel.MIP_starts.read(SOME_MST_FILE)
# you can also set parameters arbitrarily:
solver.solverModel.parameters.randomseed.set(42)
# the, you can call the solver to solve the problem
solver.callSolver(prob)
# finally, you fill the PuLP variables with the solution
status = solver.findSolutionValues(prob)
```

PR #666, opened 2023-07-05
Add option to set the number of threads for the CPLEX_PY API

### LINKED ISSUE #457: Allow for the control of the number of threads when using the python API for CPLEX
Describe the new feature
--------------------------
When creating a solver instance of CPLEX_PY I would like to be able to control the number of threads that the model uses, because some small problems that require a lot of iterations to solve benefit from using less threads.

Additional info
---------------
#### Is your feature request related to an issue? Please include the issue number.
Didn't find any

#### Does this feature exist in another product or project? Please provide a link.
The CPLEX API allows for this, as you can see in the documentation.
https://www.ibm.com/docs/en/icos/12.8.0.0?topic=parameters-global-thread-count

Implementation suggestion
---------------------------------

When creating the class CPLEX_PY allow for an optional parameter  called nthreads with a default value of 0 (also the default value of CPLEX).
When calling the function buildSolverModel at the end call a function that would run the command self.solverModel.parameters.threads.set(nthreads)

f-kretschmer:
Adapted from #528, this is a PR to resolve #457, simply adding an option to control the number of threads used by CPLEX. Please let me know, what (if anything) needs to be changed. Thanks!

### DIFF
--- a/pulp/apis/cplex_api.py
+++ b/pulp/apis/cplex_api.py
@@ -303,6 +303,7 @@ class CPLEX_PY(LpSolver):
             logPath=None,
             epgap=None,
             logfilename=None,
+            nthreads=0
         ):
             """
             :param bool mip: if False, assume LP even if integer variables
@@ -313,6 +314,7 @@ class CPLEX_PY(LpSolver):
             :param str logPath: path to the log file
             :param float epgap: deprecated for gapRel
             :param str logfilename: deprecated for logPath
+            :param int nthreads: number of threads to be used by CPLEX to solve a problem (default 0 uses all available)
             """
             if epgap is not None:
                 warnings.warn("Parameter epgap is being depreciated for gapRel")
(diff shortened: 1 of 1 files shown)


pulp/apis/cplex_api.py
```diff
@@ -303,6 +303,7 @@ def __init__(
             logPath=None,
             epgap=None,
             logfilename=None,
+            threads=0
```
pchtsp, line 306: i think we use None for the interface of other solvers. It's better to use the same defaults. See

https://github.com/coin-or/pulp/blob/d479cea0edd43eb6f017a39736738fb1008437b7/pulp/apis/cplex_api.py#L15-L40

pchtsp (changes requested):
Thanks for the PR. I left a small change to make it similar to other solvers.

pchtsp (approved):

PR #682, opened 2023-09-22
Change in highs api and status handling in both cases

### LINKED ISSUE #633: New HiGHS (highspy) interface does not set lp status
Details for the issue
--------------------

New `HIGHS` solver does not call `lp.assignStatus`.
The status is returned from:
https://github.com/coin-or/pulp/blob/master/pulp/apis/highs_api.py#L346
but in `actualSolve` the call to `lp.assignStatus` is not present.
https://github.com/coin-or/pulp/blob/master/pulp/apis/highs_api.py#L352

#### What did you do?

Solve a model using `pulp.HIGHS()`

The `lp` would have a `status` that matches the HIGHS result

`lp.status` is zero (`LpStatusNotSolved`)

Useful extra information
-------------------------

#### I installed PuLP via:

- [ ] pypi (python -m pip install pulp)
- [x] Other: latest revision: 006753076be18ee857339d333a5db11e599f21a2

#### Did you also

- [x] Searched for an existing similar issue: https://github.com/coin-or/pulp/issues?utf8=%E2%9C%93&q=is%3Aissue%20

### LINKED ISSUE #661: Inconsistent `LpStatus` with `HiGHS`
Details for the issue
--------------------

#### What did you do?
I was very interested in using `HiGHS` solver under `PuLP`, and downloaded the latest GitHub version https://github.com/coin-or/pulp/commit/a7684ac90d93dc3bca382d90841a9ab5f398b50f. I ran a small feasible problem with `HiGHS` like this:
```
import pulp 
from pulp.apis.highs_api import HiGHS
prob = pulp.LpProblem("problem", pulp.LpMaximize)
...
prob.solve(HiGHS)
```
As the problem is known to be feasible, I would expect that `prob.status = 1` since
```
pulp.LpStatus
{0: 'Not Solved', 1: 'Optimal', -1: 'Infeasible', -2: 'Unbounded', -3: 'Undefined'}
```
What I obtained is that `prob.status` is `0`, which is inconsistent with the `pulp.LpStatus` mapping. As set out in [linprog](https://docs.scipy.org/doc/scipy/reference/optimize.linprog-highs.html), `HiGHS` has a different way to represent the  exit status of the algorithm:
```
0 : Optimization terminated successfully.
1 : Iteration or time limit reached.
2 : Problem appears to be infeasible.
3 : Problem appears to be unbounded.
4 : The HiGHS solver ran into a problem.
```
The mapping with this solver must be aligned with the general `pulp.LpStatus`. 

Useful extra information
-------------------------

- [X] Linux: ( _distro:_ Ubuntu 22.04 LTS)
- [ ] Mac OS: ( _version:_ ___ )
- [ ] Other: ___

- [ ] 2.7
- [ ] 3.4
- [ ] 3.5
- [ ] 3.6
- [X] Other: 3.9

#### I installed PuLP via:

- [ ] pypi (python -m pip install pulp)
- [X] github (python -m pip install -U git+https://github.com/coin-or/pulp)
- [ ] Other: ___ (conda?)

#### Did you also

- [X] Tried out the latest github version: https://github.com/coin-or/pulp
- [ ] Searched for an existing similar issue: https://github.com/coin-or/pulp/issues?utf8=%E2%9C%93&q=is%3Aissue%20

ggsdc:
The status while solving with HiGHS was not getting updated properly and It would give back a wrong solution code on time limit but with a solution found.

This PR should fix those issues.

Should close #661 and close #633

### DIFF
--- a/pulp/apis/highs_api.py
+++ b/pulp/apis/highs_api.py
@@ -415,19 +415,19 @@ class HiGHS(LpSolver):
                     constants.LpSolutionUnbounded,
                 ),
                 HighsModelStatus.kObjectiveBound: (
-                    constants.LpStatusNotSolved,
+                    constants.LpStatusOptimal,
                     constants.LpSolutionIntegerFeasible,
                 ),
                 HighsModelStatus.kObjectiveTarget: (
-                    constants.LpStatusNotSolved,
+                    constants.LpStatusOptimal,
                     constants.LpSolutionIntegerFeasible,
                 ),
                 HighsModelStatus.kTimeLimit: (
-                    constants.LpStatusNotSolved,
+                    constants.LpStatusOptimal,
                     constants.LpSolutionIntegerFeasible,
                 ),
                 HighsModelStatus.kIterationLimit: (
-                    constants.LpStatusNotSolved,
+                    constants.LpStatusOptimal,
                     constants.LpSolutionIntegerFeasible,
                 ),
                 HighsModelStatus.kUnknown: (
--- a/pulp/apis/highs_api.py
+++ b/pulp/apis/highs_api.py
@@ -446,7 +446,7 @@ class HiGHS(LpSolver):
                 HighsModelStatus.kTimeLimit,
                 HighsModelStatus.kIterationLimit,
             ):
-                return status_dict[status][0], constants.LpSolutionNoSolutionFound
+                return constants.LpStatusNotSolved, constants.LpSolutionNoSolutionFound
             else:
                 return status_dict[status]
 
--- a/pulp/apis/highs_api.py
+++ b/pulp/apis/highs_api.py
@@ -369,39 +369,86 @@ class HiGHS(LpSolver):
             solution = lp.solverModel.getSolution()
             HighsModelStatus = highspy.HighsModelStatus
 
+            status_dict = {
+                HighsModelStatus.kNotset: (
+                    constants.LpStatusNotSolved,
+                    constants.LpSolutionNoSolutionFound,
+                ),
+                HighsModelStatus.kLoadError: (
+                    constants.LpStatusNotSolved,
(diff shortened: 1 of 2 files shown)


pchtsp (commented):
Thanks Guille. I like the mapping dictionary, because it's easy to read. Could it be kept while improving the precision that these changes bring?

ggsdc:
> Thanks Guille. I like the mapping dictionary, because it's easy to read. Could it be kept while improving the precision that these changes bring?

I've added a status mapping dict with two codes but I still have to leave an exception to the dictionary for the `kTimeLimit` and `kIterationLimit` statuses as are the only two where the actual status to give back can either be optimal + integer feasible or not solve + no solution.

ggsdc:
The checks seem to be failing due to the issue reported in #683

pchtsp:
can you please add a unit test that proves that there was something wrong and it was fixed? thanks

ggsdc:
> can you please add a unit test that proves that there was something wrong and it was fixed? thanks

I think the new test should take into account the main issue that was solved that the solver as returning status 0 on time limit even if a solution was founded. 

The new test should be able to be extended to other solvers to check correct status handling on time limit without a solution.

I will try to add (in a new PR soon) the ability for the checks, on the `pulpTestCheck` method, to also check for the solution status if given. This should allow to increase the coverage to these statuses as well.

pchtsp:
also, is this related to https://github.com/coin-or/pulp/pull/624?

ggsdc:
> also, is this related to #624?

As far as I can see point 2 of the ones mentioned on the PR should be on par with the changes done here to HiGHS.
Once we modify the tests to take into account the solution status we may find that there are instances in which thesolution status that PuLP is giving back may not be the correct one.

PR #693, opened 2023-10-02
Update bug template

ggsdc:
Changed python version on setup.py as 3.7 is no [longer receiving updates](https://peps.python.org/pep-0537/).
Also updated workflow, bug report template and deleted unused template.


ggsdc:
Maybe good to merge before merging PR #691

pchtsp:
thanks. But I would keep python 3.7 compatibility until it breaks. And as long as github actions supports testing for it. We maintained python 2.7 long after it stopped being maintained.

ggsdc:
I will just leave the bug template update then

PR #713, opened 2023-12-29
Warm Starts for HiGHS_CMD

aphi:
PR changes:

- In `highs_api.py` adds the setting of warm starts for the HiGHS_CMD solver. This involves `writesol` implementing the HiGHS raw solution format. This format doesn't have it's own documentation but rather the HiGHS devs recommend observing the format from what HiGHS outputs for its own solutions (see: https://github.com/ERGO-Code/HiGHS/issues/1284#issuecomment-1537084524).

- Adds `HiGHS_CMD` into the list of warmstart testable solvers

- Makes some unimportant typo fixes in existing code.

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -36,9 +36,9 @@ jobs:
         if: matrix.os == 'ubuntu-latest' || matrix.os == 'macOS-latest'
         run: |
           pip install highspy numpy
-      # - name: Install coptpy
-      #   run: |
-      #     pip install coptpy
+      - name: Install coptpy
+        run: |
+          pip install coptpy
       - name: Test with pulptest
         run: pulptest
       - name: Code Quality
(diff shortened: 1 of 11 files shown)


aphi:
Note: The current CI failures aren't related to this PR.

The COPT interface looks to be having trouble with test_pulp_023. If we retriggered the tests on master now they would also fail with this same error. PR #710 from dependabot is also hitting the same issue. (cc @wujianjack @pchtsp)

```ERROR: test_pulp_023 (pulp.tests.test_pulp.COPTTest) ---------------------------------------------------------------------- Traceback (most recent call last): File "/Users/runner/hostedtoolcache/Python/3.7.17/x64/lib/python3.7/site-packages/pulp/tests/test_pulp.py", line 473, in test_pulp_023 pulpTestCheck(prob, self.solver, [const.LpStatusOptimal], solution) File "/Users/runner/hostedtoolcache/Python/3.7.17/x64/lib/python3.7/site-packages/pulp/tests/test_pulp.py", line 1616, in pulpTestCheck [const.LpStatus[s] for s in okstatus], pulp.constants.PulpError: Tests failed for solver <pulp.apis.copt_api.COPT object at 0x1061f9810>: status == 0 not in [1] status == Not Solved not in ['Optimal']```

PR #718, opened 2024-01-11
Introduce the MaxNodes Option to PULP_CBC_CMD solver

dhruvk12:
We use this solver as we see the MaxNode Option is available in Parent Class, Introducing the same in the child to leverage the option that helps in providing better solutions


pulp/apis/coin_api.py
```diff
@@ -416,6 +417,7 @@ def __init__(
                 logPath=logPath,
                 mip_start=mip_start,
                 timeMode=timeMode,
+                maxNodes=None,
```
pchtsp, line 420: here you need to put `maxNodes=maxNodes`

dhruvk12: > here you need to put `maxNodes=maxNodes`

Thanks for reviewing, made the change ! @pchtsp

PR #729, opened 2024-02-02
Add SAS API interface

bolyu:
Add API interface of SAS solvers.


pulp/apis/sas_api.py
```diff
@@ -0,0 +1,623 @@
...
+            :param bool msg: if False, no log is shown
+            :param bool warmStart: if False, no warmstart or initial primal solution provided.
+            """
+            LpSolver.__init__(
+                self,
+                mip=mip,
+                msg=msg,
+            )
+            self._sas = saspy.SASsession()
+            self.warmStart=warmStart
```
pchtsp, line 126: `warmstart` can be passed to `LpSolver` and then becomes available inside `self.optionsDict`. See how it's used in other solvers. No need to assign it to the object.

pulp/apis/sas_api.py
```diff
@@ -0,0 +1,623 @@
...
+            # Only named options are allowed in SAS solvers.
+            self.solverOptions = {key:value for key, value in solverParams.items()}
+        def available(self):
```
pchtsp, line 129: I believe you need to run the black linter on the code.
https://coin-or.github.io/pulp/develop/contribute.html#applying-the-black-linter-formatter

pulp/apis/sas_api.py
```diff
@@ -0,0 +1,623 @@
...
+                except AttributeError:
+                    pass
+            data.update(self.solverOptions)
+            return data
+
+        def _create_tmp_files(self, name, *args):
+
+            return (f"{name}-pulp.{n}" for n in args)
+
+        def _delete_tmp_files(self, *args):
```
pchtsp, line 160: there already exists a `delete_tmp_files` method that does almost exactly this. No need to create a new one. The same goes for `_create_tmp_files`

pulp/apis/sas_api.py
```diff
@@ -0,0 +1,623 @@
...
+            else:
+                raise PulpSolverError("SAS94 : Objective sense should be min or max.")
+            # Get the rootnode options
+            decomp_str = self._create_statement_str("decomp")
+            decompmaster_str = self._create_statement_str("decompmaster")
+            decompmasterip_str = self._create_statement_str("decompmasterip")
+            decompsubprob_str = self._create_statement_str("decompsubprob")
+            rootnode_str = self._create_statement_str("rootnode")
+
+            if lp.isMIP():
```
pchtsp, line 215: simpler:

```
if lp.isMIP() and not self.mip:
    warnings.warn("SAS94 will solve the relaxed problem of the MILP instance.")
```

pulp/apis/sas_api.py
```diff
@@ -0,0 +1,623 @@
...
+                                        out=primalin{postfix}
+                                        dbms=csv
+                                        replace;
+                                        getnames=yes;
+                                        run;
+                                    """.format(
+                        primalin=tmpMst,
+                        postfix=postfix,
+                    )
+                    self.solverOptions["primalin"] = f"primalin{postfix}"
```
pchtsp, line 236: it's not a good practice to edit the `solverOptions` attribute. You should avoid editing during solve. It should be stateless.

pulp/apis/sas_api.py
```diff
@@ -0,0 +1,623 @@
...
+                        decompsubprob=decompsubprob_str,
+                        rootnode=rootnode_str,
+                    ),
+                    results="TEXT",
+                )
+
+            self._delete_tmp_files(tmpMps, tmpMst)
+
+            # Store log and ODS output
+            self._log = res["LOG"]
```
pchtsp, line 312: I do not see this property being defined anywhere: `_log`. Why do we need to store it inside `self`?

pulp/apis/sas_api.py
```diff
@@ -0,0 +1,623 @@
...
+            if self.msg:
+                print(self._log)
+            self._macro = dict(
```
pchtsp, line 315: same goes for `_macro`. is it needed after the solving? why?

pulp/apis/sas_api.py
```diff
@@ -0,0 +1,623 @@
...
+
+            primal_out = sas.sd2df(f"primalout{postfix}")
+            dual_out = sas.sd2df(f"dualout{postfix}")
+
+
+            if self._macro.get("STATUS", "ERROR") != "OK":
+                raise PulpSolverError("PuLP: Error ({err_name}) \
+                        while trying to solve the instance: {name}".format(
+                        err_name=self._macro.get("STATUS", "ERROR"), name=lp.name))
+            status = self._read_solution(lp, primal_out, dual_out)
```
pchtsp, line 330: what if there are no values available? should we read the soluton still?

pulp/apis/sas_api.py
```diff
@@ -0,0 +1,623 @@
...
+            return (f"{name}-pulp.{n}" for n in args)
+
+        def _delete_tmp_files(self, *args):
+            for file in args:
+                try:
+                    os.remove(file)
+                except FileNotFoundError:
+                    pass
+
+        def _write_sol(self, filename, vs):
```
pchtsp, line 424: there are many methods that are common between the two solvers. Could we just inherit one from the other and get all those methods?

pulp/apis/sas_api.py
```diff
@@ -0,0 +1,623 @@
...
+            # Get Obj Sense
+            if lp.sense == constants.LpMaximize:
+                self.solverOptions["objsense"] = "max"
+            elif lp.sense == constants.LpMinimize:
+                self.solverOptions["objsense"] = "min"
+            else:
+                raise PulpSolverError("SASCAS : Objective sense should be min or max.")
+
+            status = None
+            with redirect_stdout(SASLogWriter(self.msg)) as self._log_writer:
```
pchtsp, line 486: this part has many indentations and ifs. Can it be refactored so that it's easier to read? for example breaking it into other methods. re-using the code with different arguments, etc.

pchtsp (changes requested):
Thanks for the contribution. Please find some comments. I haven't checked everything but this should already be a good start.

I've also seen that the two python packages are available through pypi. 
1. Can you add the solver to the github CI? example: https://github.com/coin-or/pulp/blob/6e18ecbafcd3edacb34bbdc3477a6b74475891fe/.github/workflows/pythonpackage.yml#L32?
2. Can you add the solver to the testing suite? example: https://github.com/coin-or/pulp/blob/6e18ecbafcd3edacb34bbdc3477a6b74475891fe/pulp/tests/test_pulp.py#L1576?

pulp/apis/sas_api.py
bolyu: _macro is used to store some information returned by SAS solvers. It has information like primal and dual infeasibility, and SAS solution status. It might be interesting to SAS users.

pulp/apis/sas_api.py
bolyu: If no values available, the solution will be NaN.

pchtsp:
Hello, I solved the conflicts with master. Now the tests are failing because you need to run the black linter as I previously said:

https://coin-or.github.io/pulp/develop/contribute.html#applying-the-black-linter-formatter

pulp/apis/sas_api.py
```diff
@@ -0,0 +1,657 @@
...
+        LpSolver_CMD.__init__(
+            self, mip=mip, msg=msg, warmStart=warmStart, keepFiles=keepFiles
+        )
+        self.timeLimit = timeLimit
+        # Only named options are allowed in SAS solvers.
+        self.solverOptions = {key: value for key, value in solverParams.items()}
+
+    def _create_statement_str(self, statement):
+        """Helper function to create the strings for the statements of the proc OPTLP/OPTMILP code."""
+        stmt = self.solverOptions.pop(statement, None)
```
pchtsp, line 100: please do not edit this property, it should be static. There are many places where it's edited in the code.

pulp/apis/sas_api.py
```diff
@@ -0,0 +1,657 @@
...
+                return False
+            try:
+                with redirect_stdout(SASLogWriter(self.msg)) as self._log_writer:
+                    # Load the optimization action set
+                    self.cas.loadactionset("optimization")
+                    return True
+            except:
+                return False
+
+        def toDict(self):
```
pchtsp, line 466: (1) you have this function duplicated.
(2) you should not just copy the inherit method, you should call the `LpSolver_CMD.toDict` method and then add the functionality that you need.
(3) What is cas object and how do you expect to serialize it?

pulp/apis/sas_api.py
```diff
@@ -0,0 +1,657 @@
...
+                timeLimit=timeLimit,
+                **solverParams,
+            )
+            self.cas = cas
+
+        def available(self):
+            return True
+
+        def sasAvailable(self):
+            if self.cas == None:
```
pchtsp, line 456: `self.cas is None` instead

pchtsp (changes requested):
Thanks.

I've left more comments again. You need to get your code to run in the pulp CI, so fix the black linter first.

If someone else from SAS can help you reviewing the python code before asking for a review here, that would be awesome. I do not have too much time to give the amount of feedback/ improvements that is needed for this PR.

phchri:
I improved the PuLP interface with lessons learned from other interfaces, added the "with" options that other interfaces support (and a test for it) and improved performance of reading the output. @bolyu: Please review and test the changes.

bolyu:
I have reviewed and tested the changes. The updates look good to me.

pchtsp:
thanks to both. The CI tests are still failing. see below. Are you sure the code works when there's no SAS solver around?
If no SAS solver is available, the solver should return `available()=False`, which doesn't seem to be the case because the libraries are installed.

```
 SAS process has terminated unexpectedly. Pid State= (2219, 64000)

  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/pulp/apis/sas_api.py", line 233, in __init__
    self.sas = saspy.SASsession(**self._saspy_options)
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/saspy/sasbase.py", line 590, in __init__
    self._io = SASsessionSTDIO(sascfgname=self.sascfg.name, sb=self, **kwargs)
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/saspy/sasiostdio.py", line 291, in __init__
    self._startsas()
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/saspy/sasiostdio.py", line 512, in _startsas
    ll = self.submit("options svgtitle='svgtitle'; options validvarname=any validmemname=extend; ods graphics on;", "text")
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/saspy/sasiostdio.py", line 985, in submit
    raise SASIOConnectionTerminated(Exception)
saspy.sasexceptions.SASIOConnectionTerminated: No SAS process attached. SAS process has terminated unexpectedly.
<class 'Exception'>
```

phchri:
My bad, I assumed your machines would not have the SAS packages installed. I'll come up with another way of doing this next week.

phchri:
@bolyu I updated the interface classes, please review and verify the code on your side.

pulp/apis/sas_api.py
```diff
@@ -227,24 +224,21 @@ def __init__(
                 **solverParams,
             )
 
-            # Connect to saspy
+            # Try to connect to saspy, if this fails, don't throw the error here.
+            # Instead we return False on available() to be consistent with
+            # other interfaces.
             self.sas = None
             try:
                 self.sas = saspy.SASsession(**self._saspy_options)
```
bolyu, line 232: Should set saspy.logger.disabled to True when importing saspy.

pulp/apis/sas_api.py
```diff
@@ -542,33 +537,33 @@ def __init__(
...
 
         def actualSolve(self, lp):
             """Solve a well formulated lp problem"""
             log.debug("Running SAS")
 
-            if not self.sasAvailable():
-                raise PulpSolverError("""SASCAS: Cannot connect to a CAS session.""")
+            if not self.cas:
+                raise PulpSolverError(
+                    "SAS94: Cannot connect to a SAS session. Try the cfgfile option or adjust options in that file."
```
bolyu, line 564: There is a typo. It should be SASCAS in the error message.

phchri:
@bolyu I adjust the code as you suggested.

bolyu:
> @bolyu I adjust the code as you suggested.

The current version looks good.

@pchtsp Please rerun the CI tests. Thank you so much!

pchtsp (approved):

PR #745, opened 2024-04-18
handle PL bounds and include test

### LINKED ISSUE #635: PuLP does not recognize PL bound type in .mps files
**System:**
MacBook (INTEL CPU), Ventura 13.2, Python 3.8
PuLP 2.7.0

From Wikipedia : 
[https://en.wikipedia.org/wiki/MPS_(format)](url)
`The optional BOUNDS section specifies lower and upper bounds on individual variables, if they are not given by rows in the matrix.  .... Bound type PL is for a free positive from zero to plus infinity, but as this is the normal default, it is seldom used.`

About 10% of models in MipLib2017 benchmark contain `PL Bound  variable` lines and PuLP throws an exception when running 
`var, prob = LpProblem.fromMPS(modelName)`

The fix appears to be easy: in mps_lp.py, under `readMPSSetBounds` add two lines.  There is no need to call `set_both_bounds` because the bounds are <0, infinity> which is the default

```
   if bound == "FR":
        set_both_bounds(None, None)
        return
    elif bound == "BV":
        set_both_bounds(0, 1)
        return
    # add these two lines
    elif bound == "PL":
        return
```

rutger-van-beek-cqm:
Fixes #635. 

Some software writes MPS-files including PL bounds, but without a value. When reading these lines the `readMPSSetBounds` raises an `IndexError`, when trying to read the value. 

PL-bounds can be ignored. This PR implements a case in the `readMPSSetBounds` to ignore the PL-type bounds. 

Next to that a test is added, to test whether a MPS-file containing a PL-bound can be read.

### DIFF
--- a/pulp/mps_lp.py
+++ b/pulp/mps_lp.py
@@ -175,6 +175,9 @@ def readMPSSetBounds(line, variable_dict):
     elif bound == "BV":
         set_both_bounds(0, 1)
         return
+    elif bound == "PL":
+        # bounds equal to defaults
+        return
     value = float(line[3])
     if bound in ["LO", "UP"]:
         set_one_bound(bound, value)
--- a/pulp/tests/test_pulp.py
+++ b/pulp/tests/test_pulp.py
@@ -42,6 +42,10 @@ BOUNDS
 ENDATA
 """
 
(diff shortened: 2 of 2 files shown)


rutger-van-beek-cqm:
I will make a new PR where the commit can be linked to my account.

Issue #749, opened 2024-05-03
Slow performance instantiating large problems


MBradbury:
Details for the issue
--------------------

#### What did you do?

Creating large problems in PULP is slow, there are a two issues:
1. Creating expressions via `pulp.lpSum` is slower than directly creating them via `pulp.LpAffineExpression`
2. Making constraints requires allocating and filling an additional 2 copies of an LpAffineExpression

```python
import pulp, numpy as np, itertools

# Some reward matrix
reward = np.full((300, 200, 50), 10)

prob = pulp.LpProblem("Problem", pulp.constants.LpMaximize)
dv = pulp.LpVariable.dicts("dv",
       itertools.product(range(300), range(200), range(50)),
       cat=pulp.constants.LpBinary)
```

Slow objective function:
```python
prob += pulp.lpSum([
            dv[a, b, c] * reward[a, b, c]
            for a in range(300)
            for b in range(200)
            for c in range(50)
        ])
```

Faster objective function:
```python
prob += pulp.LpAffineExpression(
            (dv[a, b, c], reward[a, b, c])
            for a in range(300)
            for b in range(200)
            for c in range(50)
        )
```

Slow constraint:
```python
for c in range(50):
  prob += pulp.lpSum([
              dv[a, b, c]
              for a in range(300)
              for b in range(200)
          ]) == 1
```

Faster constraint:
```python
for c in range(50):
  prob += pulp.LpAffineExpression(
              (dv[a, b, c], 1)
              for a in range(300)
              for b in range(200)
          ) == 1
```

Making constraints is slow as `self - other` creates the first LpAffineExpression copy at https://github.com/coin-or/pulp/blob/master/pulp/pulp.py#L980, https://github.com/coin-or/pulp/blob/master/pulp/pulp.py#L983 or https://github.com/coin-or/pulp/blob/master/pulp/pulp.py#L986. A second copy is then made when LpConstraint's constructor is called (https://github.com/coin-or/pulp/blob/master/pulp/pulp.py#L1012).

It seems as if one of these copies can be avoided if the `rhs` is set when `other` is int/float. So in `LpAffineExpression.__eq__` this would be:
```python
def __eq__(self, other):
        if isinstance(other, (float, int)):
            return LpConstraint(self, const.LpConstraintEQ, rhs=other)
        else:
            return LpConstraint(self - other, const.LpConstraintEQ)
```

[...]

PR #750, opened 2024-05-04
Improve performance for large models

MBradbury:
Addresses some performance issues identified in #749.

Primary changes:
1. LpConstraint no longer inherits from LpAffineExpression, this is so a reference can be kept to the already created object instead of copying it
2. Avoid creating temporary LpAffineExpression when building a LpConstraint with a RHS that is an int/float
3. Avoid creating temporary lists (e.g., of Iterables before iterating over them, before transforming lists into another form)

In some experiments with a model I am working with I have observed 4x improvement in setting constraints. pulp/apis/cplex_api.py:buildSolverModel has gone from 70s to 50s.

Comparison of numbers in performance script from #749 comparing the original performance to performance with this PR. Using list of tuples and passing that directly to LpAffineExpression remains faster than using lpSum.

| Name      | Original      | PR |
| ------------- | ------------- | ------------- |
| Slow Objective | 16.692 | 16.595 |
| Slow Constraint | 22.000 | 6.435 |
| Fast Objective | 2.945 | 2.999 |
| Fast Constraint | 20.278 | 5.552 |


MBradbury:
Just a note that I added some type hinting to help me understand the code. Happy to remove if needed.

pchtsp:
please run the black linter as per instructions: https://coin-or.github.io/pulp/develop/contribute.html#applying-the-black-linter-formatter (and check any other step you may be missing). I'll try to review the changes soon.

pulp/apis/cplex_api.py
```diff
@@ -339,6 +339,8 @@ def buildSolverModel(self, lp):
                 lp.solverModel.objective.set_sense(
                     lp.solverModel.objective.sense.maximize
                 )
+            if lp.objective is None:
+                raise PulpSolverError("No objective set")
```
pchtsp, line 343: in theory it's possible to not set an objective function afaik and just model a constraint satisfaction problem. Isn't that the case?

pchtsp (commented):
Thanks for the changes. they look comprehensive. I also appreciate the type hints.

I'm curious though if there was a way to keep the inheritance of `LpAffineExpression` without making a copy when initializing the `LpConstraint`. Can't we just modifiy the `LpConstraint` init in line 1012 to fill the values without copying them? and then we would keep everything else as is.

MBradbury:
> Thanks for the changes. they look comprehensive. I also appreciate the type hints.
> 
> I'm curious though if there was a way to keep the inheritance of `LpAffineExpression` without making a copy when initializing the `LpConstraint`. Can't we just modifiy the `LpConstraint` init in line 1012 to fill the values without copying them? and then we would keep everything else as is.

This will still lead to a large copy as it is essentially copying a dict to another dict. So for the test file that would be a copy of a 3 million entry dict.

Another approach could be to have LpAffineExpression not inherit from dict and instead have it as a variable. Then the constraint would just use that variable when inherited from LpAffineExpression, avoiding the large copy.

MBradbury:
I have tried to decouple LpAffineExpression from its dict base class, but this has led to test failures I cannot get to the bottom of. This commit can be found here: https://github.com/MBradbury/pulp/commit/12feb605e9b84af2c030d1ffb25e13110ba94100

stumitchell:
Looking good however, I suggest that we do a major version change as the change to modifying things in place instead of copying may cause errors in existing code.

pchtsp:
it seems some tests are failing

MBradbury:
Looks like `LpConstraint` needs a `asCplexLpAffineExpression` method that calls the method of the same name on `expr`. Or the test can be updated to use `LpProblem.objective.expr.asCplexLpAffineExpression` instead of `LpProblem.objective.asCplexLpAffineExpression`.

christiansegercrantz:
What's the current situation with this PR, is it still being worked on? I'm currently experiencing a very similar situation where the solving itself is taking 1.7s but I accumulate 22s of time from LpSum.

MBradbury:
There was a small change that I mentioned in an earlier message, I just don't have the time to address it currently. I expect that this will be an issue for me in March/April, when I need to use Pulp for an annual task, so will likely take a look again then.

When I originally took a look the biggest design issue is that `LpAffineExpression` inherits from a dict. Ideally this would not be the case to avoid excessive copying, but fixing this issue was even harder to get right than the current attempt at speeding up Pulp.

christiansegercrantz:
Looks like you've made good progress. Are you planning the continue the work still or can we expect the work to be merged soon?

christiansegercrantz:
Alright, so we are waiting for maintainer approval and release then?

pchtsp:
thanks! can you please apply the black linter as per instructions here: https://coin-or.github.io/pulp/develop/contribute.html#applying-the-black-linter-formatter

christiansegercrantz:
Great job! @pchtsp When can we expect a release with this?

pchtsp:
@christiansegercrantz 
following @stumitchell advice above, I just release PuLP 3.0.0 to pypi with these changes.

Issue #650, opened 2023-05-11
static type checks with `mypy`


AngryMaciek:
Describe the new feature
--------------------------

Is there any way to build/install package stubs for _pulp_?  
I was running static type check analysis with `mypy` but these seem missing:

```
file.py:19: error: Skipping analyzing "pulp": module is installed, but missing library stubs or py.typed marker  [import]
```

Additional info
---------------
Please answer these questions before submitting your feature request.

#### Is your feature request related to an issue? Please include the issue number.

No, I have checked other issues, nothing seems related to this right now.

#### Does this feature exist in another product or project? Please provide a link.

Some project already have type stubs included inside (ex: `numpy`), for some others there are external packages, like `pandas`: https://pypi.org/project/pandas-stubs/

CubeSugarCheese:
Is there any latest progress on this issue?
The infomation about type hint are here: https://peps.python.org/pep-0484/ and https://docs.python.org/3/library/typing.html
Before adding type hints, we need to answer some questions.

1. Embedded or External?
We can use external type hint (in *.pyi file), it will add some maintenance costs to handle seperate declarations.
Or we can directly add to py files, it will force subsequent contributors to handle it.

2. what's the lowest compatible python version?
Type hint having gone through multiple python versions to improve.
For example, `ParamSpec` and `Concatenate` added in python 3.10, they are use for decorator,
and we can use `A | B` instead of `typing.Union[A, B]`.
Python 3.11 added TypeVarTuple for variadic generic support.
Python 3.9 added builtin collection subscripting, we can use `list[int]` instead of `typing.List[int]`.

Note that readme described "PuLP requires Python 3.7 or newer.".
The target verion is 3.7 ?

CubeSugarCheese:
> Is there any latest progress on this issue? The infomation about type hint are here: https://peps.python.org/pep-0484/ and https://docs.python.org/3/library/typing.html Before adding type hints, we need to answer some questions.
> 
> 1. Embedded or External?
>    We can use external type hint (in *.pyi file), it will add some maintenance costs to handle seperate declarations.
>    Or we can directly add to py files, it will force subsequent contributors to handle it.
> 2. what's the lowest compatible python version?
>    Type hint having gone through multiple python versions to improve.
>    For example, `ParamSpec` and `Concatenate` added in python 3.10, they are use for decorator,
>    and we can use `A | B` instead of `typing.Union[A, B]`.
>    Python 3.11 added TypeVarTuple for variadic generic support.
[...]

PR #755, opened 2024-05-29
add type hints (Experimental)

CubeSugarCheese:
#650 

This pr still incomplete, some functions are hard to type, like `LpVariable.matrix` and `LpVariable.dicts`.
And lots of usage of `cast` and `# type: ignore` to adapt old code base (some replace by not None assert).

### DIFF
--- a/pulp/apis/core.py
+++ b/pulp/apis/core.py
@@ -30,17 +30,34 @@ Note that the solvers that require a compiled extension may not work in
 the current version
 """
 
+from typing import (
+    Callable,
+    TypeVar,
+    Union,
+    Optional,
+    Tuple,
+    Any,
+    List,
+    Dict,
+    Type,
+    TYPE_CHECKING,
+)
+
+if TYPE_CHECKING:
+    from pulp.pulp import LpProblem, LpVariable
+
 import os
 import platform
 import shutil
 import sys
 import ctypes
+from ctypes import Array, c_char_p, c_double, c_char
 
 
 from time import monotonic as clock
 
 import configparser
-from typing import Union
+
 
 Parser = configparser.ConfigParser
 
@@ -49,17 +66,20 @@ from .. import constants as const
 
 import logging
 
-try:
-    import ujson as json
-except ImportError:
+if TYPE_CHECKING:
     import json
+else:
+    try:
+        import ujson as json
+    except ImportError:
+        import json
 
 log = logging.getLogger(__name__)
 
 import subprocess
 
 devnull = subprocess.DEVNULL
-to_string = lambda _obj: str(_obj).encode()
+to_string: Callable[[Any], bytes] = lambda _obj: str(_obj).encode()
 
 from uuid import uuid4
 
@@ -73,7 +93,9 @@ class PulpSolverError(const.PulpError):
 
 
 # import configuration information
-def initialize(filename, operating_system="linux", arch="64"):
+def initialize(
+    filename: str, operating_system: str = "linux", arch: str = "64"
+) -> Tuple[Any, ...]:
     """reads the configuration file to initialise the module"""
     here = os.path.dirname(filename)
     config = Parser({"here": here, "os": operating_system, "arch": arch})
@@ -160,14 +182,14 @@ else:
 operating_system = None
 if sys.platform in ["win32", "cli"]:
     operating_system = "win"
-    PULPCFGFILE += ".win"
+    PULPCFGFILE += ".win"  # type: ignore
 elif sys.platform in ["darwin"]:
     operating_system = "osx"
     arch = "64"
-    PULPCFGFILE += ".osx"
+    PULPCFGFILE += ".osx"  # type: ignore
 else:
     operating_system = "linux"
-    PULPCFGFILE += ".linux"
+    PULPCFGFILE += ".linux"  # type: ignore
 
 DIRNAME = os.path.dirname(__file__)
 config_filename = os.path.normpath(os.path.join(DIRNAME, "..", PULPCFGFILE))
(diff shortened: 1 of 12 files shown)


pulp/pulp.py
```diff
@@ -880,38 +1003,145 @@ def addInPlace(self, other, sign=1):
...
+                Union["LpAffineExpression", LpVariable, "LpConstraintVar", int, float]
+            ],
+            Iterable[LpVariable],
+            LpVariable,
+            "LpConstraintVar",
+            int,
+            float,
+            None,
+        ],
+    ) -> "LpAffineExpression":
```
pchtsp, line 1021: this `other` is repeated in several places. (1) I'm not sure it's correct (2) if it is, we should create it once and re-use it

pulp/pulp.py
```diff
@@ -1302,15 +1755,14 @@ def __init__(self, name="NoName", sense=const.LpMinimize):
         if " " in name:
             warnings.warn("Spaces are not permitted in the name. Converted to '_'")
             name = name.replace(" ", "_")
-        self.objective = None
-        self.constraints = _DICT_TYPE()
+        self.constraints = dict()
```
pchtsp, line 1758: why the change? it's the same I believe

pulp/pulp.py
```diff
@@ -1302,15 +1755,14 @@ def __init__(self, name="NoName", sense=const.LpMinimize):
...
         self.name = name
         self.sense = sense
         self.sos1 = {}
         self.sos2 = {}
         self.status = const.LpStatusNotSolved
         self.sol_status = const.LpSolutionNoSolutionFound
-        self.noOverlap = 1
+        self.noOverlap = False
```
pchtsp, line 1765: this changes the code. Shouldn't it be True?

pulp/pulp.py
```diff
@@ -1322,9 +1774,9 @@ def __init__(self, name="NoName", sense=const.LpMinimize):
         self.solutionCpuTime = 0
 
         # locals
-        self.lastUnused = 0
+        self.lastUnused = True
```
pchtsp, line 1777: False?

pulp/pulp.py
```diff
@@ -1434,13 +1890,12 @@ def fromDict(cls, _dict):
         # objective function.
         # we change the names for the objects:
         obj_e = {var[v["name"]]: v["value"] for v in _dict["objective"]["coefficients"]}
-        pb += LpAffineExpression(e=obj_e, name=_dict["objective"]["name"])
+        pb += LpAffineExpression(e=obj_e, name=_dict["objective"]["name"])  # type: ignore
 
         # constraints
         # we change the names for the objects:
-        def edit_const(const):
-            const = dict(const)
```
pchtsp, line 1442: why are you taking out the shallow copy?

pchtsp (commented):
sorry for the delay.
I've checked around half.

Many things look good. Many others require careful investigation. I'm particularly skeptic on the large Union types you have for some of the `other` arguments.

In some places you're changing the behavior of the code (True instead of 0)

You use `cast`ing a lot. I'm not sure if it's really required if we're correctly defining the types. Also, there are some `ignore` that I believe should not be needed with the correct typing.

You're also cleaning a lot of compatibility code (which is fine) but I'm not sure if that's breaking something. Could you at least comment on the pieces of code you're taking out?

pulp/apis/core.py
```diff
@@ -238,15 +273,43 @@ def copy(self):
...
+        Any,
+        "Array[c_double]",
+        "Array[c_double]",
+        "Array[c_double]",
+        "Array[c_char_p]",
+        "Array[c_char_p]",
+        "Array[c_char]",
+        Dict[int, "LpVariable"],
+        Dict[Any, Any],
+    ]:
```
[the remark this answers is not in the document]
CubeSugarCheese: I think ok. we can define a type alias in other file and import to use at this.

pulp/apis/core.py
```diff
@@ -213,23 +248,23 @@ def __init__(
...
+    def actualResolve(self, lp: "LpProblem", **kwargs: Any):
         """
         uses existing problem information and solves the problem
         If it is not implemented in the solver
         just solve again
         """
         self.actualSolve(lp, **kwargs)
 
-    def copy(self):
+    def copy(self) -> "Self":
```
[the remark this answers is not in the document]
CubeSugarCheese: yes, it should return "LpSolver". I will fix it.

pulp/pulp.py
CubeSugarCheese: Sorry, it should be True.

CubeSugarCheese:
Thanks for your review.

At first, I think I can typed this lib suitably. But after some research, I'm got a discouraged.
There many imprecision in old code, for example.

1. Some place use implicit cast to bool to check None, it cant be understand by typechecker. And many places use literal 1 and 0 as bool, and some place use bool literal in the same time. they all come from [803c3d3], a 12 years ago commit. I dont know why, after search this keyword in all place just use as bool, I replace it as bool.
2. Some place check None and some place not check in the same object/attr, even they init by None.
3. Some recursive algorithm hard to typed.
4. LpElement is inherited by LpVariable and LpConstraintVar, but almost all code assume a LpElement object must be LpVariable or LpConstraintVar, it violating the rules of the subtype system, so all LpElement replaced by Union[LpVariable, LpConstraintVar].

Sorry, the workload for this task far exceeds my estimation. In contrast, the benefits obtained from typed are too low. Now I no longer have courage and energy to modify these codes.
Perhaps the only feasible way to type this lib is to start writing from scratch. So that we can let go of the burden of compatibility.

You can close this pr.
Sorry again.

PR #769, opened 2024-08-21
add test / fix for lpdot

frrad:
Without the change to `lpdot`, this new test fails with 

```
E       AssertionError: assert [{'name': 'x', 'value': 1}] == [{'name': 'x', 'value': 2}]
E         
E         At index 0 diff: {'name': 'x', 'value': 1} != {'name': 'x', 'value': 2}
E         Use -v to get more diff
```

The `lpdot` function uses `isiterable` to check for "vector" types, however an `LpAffineExpression` is iterable.

### DIFF
--- a/pulp/constants.py
+++ b/pulp/constants.py
@@ -87,15 +87,6 @@ LpConstraintSenses = {LpConstraintEQ: "=", LpConstraintLE: "<=", LpConstraintGE:
 LpCplexLPLineSize = 78
 
 
-def isiterable(obj):
-    try:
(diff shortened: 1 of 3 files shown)


frrad:
I did a bit more looking around and found https://github.com/coin-or/pulp/pull/572 and https://github.com/coin-or/pulp/issues/483 for the same issue :-/

Just wanted to flag so that when you eventually take one of them 🤞 you can close the other.

frrad:
Thanks!

PR #772, opened 2024-08-31
refactor "test" to a test

frrad:
there was some dangling test code, but no actual test that exercised the Matrix class.


pchtsp:
hello! and thanks. Could you add it importing unittests like the test_pulp.py file example? Thay way they're all tested together.

frrad:
Done! It also seemed to be necessary to register the test in `run_tests.py`. Let me know if that was not the right thing to do.

pchtsp (approved):

PR #773, opened 2024-09-02
type `sparse.py`, enable mypy in CI

frrad:
This is on top of https://github.com/coin-or/pulp/pull/772 - the diff will be a little cleaner once that is merged.

This PR:
1) adds type hints for `sparse.py`. I have made no functional changes, only types.
1) adds a `pyproject.toml` which blacklists all files which don't currently pass mypy (this is most of them)
1) changes `.github/workflows/pythonpackage.yml` to run mypy
1) removes the (I believe) unused `.travis.yml` config


pchtsp:
hello! thanks. Why the pyproject.toml? and why type hint sparse? also, could you check https://github.com/coin-or/pulp/pull/755 as this was the current work for the type hinting.

frrad:
Hello!

I have added `pyproject.toml` in order to configure `mypy`. The current configuration is such that `mypy .` will pass. In particular, the addition I made to your `.gihub/...` configuration file should successfully run mypy on future PRs.

By introducing this check, we ensure that future files are mypy'd from the start, and give ourselves a nice way to ratchet up the type coverage - you can do smaller PRs which remove one or two files from `pyproject.toml` and add types to them. I chose `sparse.py` because it was an easy starting point.

I did notice https://github.com/coin-or/pulp/pull/755 

1. since `sparse.py` is so isolated you should just be able to "take theirs" when you merge this branch into it and have no conflicts
2. I think that PR is a compelling piece of evidence in favor of the incremental approach - instead of having a years-long back and forth over one monolithic change you can just take small improvements as they come with less back and forth.

requirements-dev.txt
```diff
@@ -1,4 +1,5 @@
+black==23.3.0
+mypy==1.4.1
```
pchtsp, line 2: please leave the versions open. That is:

```
black
mypy
```

pyproject.toml
```diff
@@ -0,0 +1,70 @@
+[tool.mypy]
```
pchtsp, line 1: Can you just add a directory instead of each file? something like seems to work according to https://stackoverflow.com/a/78480102/6508131:

```python
exclude = [
"examples/",
"doc",
"pulp/apis",
"pulp/solverdir",
]
```

pchtsp (commented):
Thanks, I've left a couple more comments and we should merge afterwards.

PR #778, opened 2024-10-02
Fix pulp cbc cmd solver params

marcusreaiche:
PR

- The PULP_CBC_CMD solver is not correctly processing some parameters when calling CBC.
- This is the case for the presolve and cuts arguments.
- For example, when calling solve with `presolve=False`, the option `presolve on` is still passed to the command line. 
- Moreover, Pulp's documentation on PULP_CBC_CMD states that the strong parameter is of type bool.
- However, it is of type int as mentioned is CBC's help.
- New unit tests that check the command line instruction passed to CBC are created.

The following Google Colab Notebook presents all the issues and the corresponding fixes.

https://colab.research.google.com/drive/18afePVGohf9Bumns6ziiXjuoAI6SrJZL?usp=sharing


marcusreaiche:
Commit 29ab059 removed the type annotations for compatibility with older Python versions.

pchtsp:
Thanks, until now, we had them as flags of None/ not None. I guess it makes sense to be able to pass False if we want to deactivate them. 

Since these tests only affect CBC, I recommend to add them as methods of the CBC class:

https://github.com/coin-or/pulp/blob/6af38012df67a9d8327593d4c8f8a1b5649d59c7/pulp/tests/test_pulp.py#L1439-L1443

marcusreaiche:
Hey Franco,

Thanks for your recommendations.

- I moved the new tests to the PULP_CBC_CMDTest.
- I also moved the helper methods implemented in pulp/tests/utilities.py to the same class as static methods
    - Since they are only used to extract information from the CBC log file, I believe it is more logical to include them within this class.

Please let me know if you have any other suggestions.

pchtsp (approved):

PR #784, opened 2024-11-02
Modified example/wedding.py and related files to maximize happiness

aralsea:
This PR is a simple fix in example/wedding.py. Since the objective function is defined as `happiness`, it makes more sense to use maximize rather than minimize.

### DIFF
--- a/doc/KPyCon2009/code/wedding.py
+++ b/doc/KPyCon2009/code/wedding.py
@@ -27,7 +27,7 @@ x = pulp.LpVariable.dicts(
     "table", possible_tables, lowBound=0, upBound=1, cat=pulp.LpInteger
 )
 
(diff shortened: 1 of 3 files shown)


pchtsp:
I think you then have to change the returned value of the happiness function, to keep it working as before. For exmple by multiplying by minus 1 or making the inverse of the value

aralsea:
@pchtsp Thank you for your quick feedback! 
Based on the code code content, I initially thought that the definition of the happiness function seems to have been chosen somewhat arbitrarily, so I didn't make the inverse. If you think an adjustment would improve the function’s consistency, I’d be happy to make the change.

aralsea:
In addition, I also added similar changes to related files that I initially missed.

pchtsp:
I'm closing the issue for now. I think the happiness function should return the negative of the current value. Because the model currently minimizes the cost, not maximizes the happiness.

PR #787, opened 2024-11-13
add license to setup.py

### LINKED ISSUE #786: License in PyPI and GitHub is different
I noticed a discrepancy regarding the license for PuLP. On PyPI, it’s listed under a BSD license, whereas the GitHub repository states that it’s under an MIT license. This difference might lead to some confusion for users. If possible, aligning the license information across both platforms would be very helpful.

michirakara:
this PR should fix #786

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,6 +19,7 @@ with open(readme_name) as fh:
 setup(
     name="PuLP",
     version=VERSION,
+    license="MIT",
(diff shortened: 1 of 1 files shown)


pchtsp (commented):
Thanks! also, on the classifiers further below, it says `"License :: OSI Approved :: BSD License",` I guess we should change that too? Not sure to what though.

pchtsp (approved):

Issue #791, opened 2025-02-12
Unsupported bound type `MI` in `pulp.LpProblem.fromMPS`


dwr-zroy:
Details for the issue
--------------------

#### What did you do?
I attempted to read an `.mps` file using `pulp.LpProblem.fromMPS`.

I discovered the issue when converting problems to `pulp` style `.json` files, and an `.mps` file containing an `ML` type bound caused `pulp` to error. `ML` bounds specify that the lower bound is `-inf` for a variable ([see note E here](https://lpsolve.sourceforge.net/5.5/mps-format.htm))

##### MPS File bounds section (simplified):

```
BOUNDS
...
 MI bnd    d_losvq_trnsf                     
 UP bnd    d_losvq_trnsf    400
...
ENDATA
```

#### What did you expect to see?
Successful file read.

#### What did you see instead?
`IndexError` raised.

In the function `readMPSSetBounds`, a series of checks are made against the bound type. Bound types that do not specify a bound value are caught and handled separately (currently `FR`, `BV`, and `PL` are caught). If the type is not caught, the bound value is read from the `lines` object, which is a list.

Since the `MI` bound type does not specify a bound value, and it is not caught by the if statements, an index error is raised since the `lines` list is an unexpected length.

Proposed Solution
-------------------------

Adding an additional clause to `readMPSSetBounds` in `pulp.mps_lp`.

```python
def readMPSSetBounds(line, variable_dict):
    bound = line[0]
    var_name = line[2]

    def set_one_bound(bound_type, value):
        variable_dict[var_name][BOUNDS_EQUIV[bound_type]] = value

    def set_both_bounds(value_low, value_up):
        set_one_bound("LO", value_low)
        set_one_bound("UP", value_up)

    if bound == "FR":
        set_both_bounds(None, None)
        return
    elif bound == "BV":
        set_both_bounds(0, 1)
        return
    elif bound == "PL":
        # bounds equal to defaults
        return
    elif bound == "MI":
        # Lower bound -inf
        set_one_bound("LO", None)
        return

    value = float(line[3])
    if bound in ["LO", "UP"]:
        set_one_bound(bound, value)
    elif bound == "FX":
        set_both_bounds(value, value)
    return
```

I am uncertain if this causes side effects, as I have not tested the solution against the unit-tests myself. I have successfully used this as a "patch" to finish my conversion effort.

[...]

PR #792, opened 2025-02-13
Add support for bound type `MI` in `pulp.LpProblem.fromMPS`

dwr-zroy:
Adds support for `MI` bounded variables in the `pulp.LpProblem.fromMPS` method . 

This addresses Issue #791.

Changes
------------

Added an additional clause to `readMPSSetBounds` in `pulp.mps_lp`.

```
def readMPSSetBounds(line, variable_dict):
    bound = line[0]
    var_name = line[2]

    def set_one_bound(bound_type, value):
        variable_dict[var_name][BOUNDS_EQUIV[bound_type]] = value

    def set_both_bounds(value_low, value_up):
        set_one_bound("LO", value_low)
        set_one_bound("UP", value_up)

    if bound == "FR":
        set_both_bounds(None, None)
        return
    elif bound == "BV":
        set_both_bounds(0, 1)
        return
    elif bound == "PL":
        # bounds equal to defaults
        return
    elif bound == "MI":
        # Lower bound -inf
        set_one_bound("LO", None)
        return

    value = float(line[3])
    if bound in ["LO", "UP"]:
        set_one_bound(bound, value)
    elif bound == "FX":
        set_both_bounds(value, value)
    return
```

Added a test to `BaseSolverTest.PuLPTest` called `test_importMPF_MI_bound`, which is similar in form to the existing `test_importMPS_PL_bound` method. An additional global variable is created, called `EXAMPLE_MPS_MI_BOUNDS` which is similar in form to the existing `EXAMPLE_MPS_PL_BOUNDS` variable.

Tests pass on Windows 11, with the `PULP_CBC_CMD` solver. Additional testing with more solvers is requested.

Documentation was re-built, but no changes are present.

`isort` also cleaned up the order of imports in each of the changed files. Let me know if this was inappropriate. 

Next Steps
------------

Testing with additional solvers should be considered by those more familiar with the project. I do not have licensing for most of the supported solvers so that testing was not done on my part. However, this change doesn't touch any of the library related to the solver APIs.

### DIFF
--- a/pulp/mps_lp.py
+++ b/pulp/mps_lp.py
@@ -5,6 +5,7 @@
 """
 
 import re
+
 from . import constants as const
 
 CORE_FILE_ROW_MODE = "ROWS"
@@ -177,6 +178,9 @@ def readMPSSetBounds(line, variable_dict):
         return
     elif bound == "PL":
         # bounds equal to defaults
+        return
+    elif bound == "MI":
+        set_one_bound("LO", None)
         return
 
     value = float(line[3])
--- a/pulp/tests/test_pulp.py
+++ b/pulp/tests/test_pulp.py
@@ -2,18 +2,19 @@
 Tests for pulp
 """
 
+import functools
 import os
+import re
 import tempfile
+import unittest
 
-from pulp.constants import PulpError
-from pulp.apis import *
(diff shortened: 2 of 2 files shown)


dwr-zroy:
Just as a note, I think the CLA bot above has commented in the wrong order. I have since signed the CLA.

Issue #794, opened 2025-02-20
Bug Report: Incorrect Sum Constraint Generation in PuLP v3.0


hanapirath:
Details for the issue
--------------------

When using lpSum to define a sum constraint and applying both Min and Max constraints to the resulting variable, PuLP generates the constraints incorrectly. Specifically, the Min_Constraint is overwritten or altered when the Max_Constraint is added. This issue occurs when multiple constraints are applied sequentially to the sum of variables. See example below.

```
import sys
import pulp

print(f"Python version: {sys.version}")
print(f"pulp version: {pulp.__version__}")

# Create a minimization problem
model = pulp.LpProblem("Sum_Constraint_Example", pulp.LpMinimize)

# Define individual variables x[i] for i in range(3)
x_vars = {i: pulp.LpVariable(f"x{i}", lowBound=0, cat=pulp.LpContinuous) for i in range(3)}

# Define x as the sum of all x[i]
x = pulp.lpSum(x_vars[i] for i in range(3))

# Add min and max constraints on x
model += (x >= 2, "Min_Constraint")
model += (x <= 5, "Max_Constraint")  

# Print model
print(model)
```

#### What Pulp generated

> Python version: 3.10.12 (main, Sep 11 2024, 15:47:36) [GCC 11.4.0]
> pulp version: 3.0.0
> Sum_Constraint_Example:
> MINIMIZE
> None
> SUBJECT TO
> Min_Constraint: x0 + x1 + x2 >= 7
> 
> Max_Constraint: x0 + x1 + x2 <= 7
> 
> VARIABLES
> x0 Continuous
> x1 Continuous
> x2 Continuous

Useful extra information
-------------------------

The info below often helps, please fill it out if you're able to. :)

#### What operating system are you using?

- [ ] Windows: ( _version:_ ___ )
- [ ] Linux: ( _distro:_ ___ )
- [ ] Mac OS: ( _version:_ ___ )
- [ ] Other: ___

#### I'm using python version:

- [ ] 3.7
- [ ] 3.8
- [ ] 3.9
- [x] 3.10
- [ ] 3.11
- [ ] Other: ___

#### I installed PuLP via:

- [x] pypi (python -m pip install pulp)
- [ ] github (python -m pip install -U git+https://github.com/coin-or/pulp)
- [ ] Other: ___ (conda?)

#### Did you also

- [ ] Tried out the latest github version: https://github.com/coin-or/pulp
[...]

PR #796, opened 2025-02-20
Add tests for operators on LpConstraint

MBradbury:
Following on from https://github.com/coin-or/pulp/issues/794#issuecomment-2671127824. Add tests for add/subtract/copy/negate/multiply/divide on `LpConstraint`.

Some functionality has been removed. For example, multiplying / dividing a LpConstraint by a LpConstraint, as there does not seem to be a sensible way for this to have a linear result.

### DIFF
--- a/pulp/pulp.py
+++ b/pulp/pulp.py
@@ -128,7 +128,7 @@ import sys
 import warnings
 import math
 from time import time
-from typing import Any
+from typing import Any, Literal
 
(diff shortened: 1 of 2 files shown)


pchtsp:
ok, published v3.0.2, fingers crossed!

MBradbury:
Indeed! Apologies for all the breakages.

Issue #793, opened 2025-02-19
Bug: Execution hangs after running HiGHS solve.


christiansegercrantz:
Details for the issue
--------------------

#### What did you do?
I tried running the HiGHS solver (using `pulp.apis.HiGHS`, highspy == 1.9.0 (fa40bdf)) on my project with the latest version of PulP from Github. The execution hangs after the solver has completed its work and never progresses forward in the code. I am unsure what information to provide to easier debug it, so please do ask for more. The solver used the following parameters:

```python
solver = pulp.apis.HiGHS(
     gapAbs=100,
     timeLimit=60 * 5,
     threads=6
)
model.solve(solver)
```

#### What did you expect to see?
That the execution is completed and the `model.solve`operation completed. 

#### What did you see instead?
The solver hangs indefinately.

Useful extra information
-------------------------

The info below often helps, please fill it out if you're able to. :)

#### What operating system are you using?

- [ ] Windows: ( _version:_ ___ )
- [ ] Linux: ( _distro:_ ___ )
- [X] Mac OS: macOS Sequoia 15.3 (24D60)
- [ ] Other: ___

#### I'm using python version:

- [ ] 3.7
- [ ] 3.8
- [ ] 3.9
- [ ] 3.10
- [ ] 3.11
- [X] Other: cpython@3.12.8

#### I installed PuLP via:

- [ ] pypi (python -m pip install pulp)
- [ ] github (python -m pip install -U git+https://github.com/coin-or/pulp)
- [X] Other: `rye add pulp --git https://github.com/coin-or/pulp`

#### Did you also

- [X] Tried out the latest github version: https://github.com/coin-or/pulp
- [X] Searched for an existing similar issue: https://github.com/coin-or/pulp/issues?utf8=%E2%9C%93&q=is%3Aissue%20

pchtsp:
can you share the model in some of the supported formats?
https://coin-or.github.io/pulp/guides/how_to_export_models.html

If it's a very big model, then it may be something wrong with post-processing, or memory, or anything, really.
Does it work correctly with a small instance of the model?

Also, if you put a shorter timeLimit, does it work? Do you see any log or output from the solver?

christiansegercrantz:
I doubt it's the size since It takes approx 3 seconds to solve. Additionally, the solver prints the solving report indicating it's ready:
<details>
<summary>Solving report</summary>
Solving report
  Status            Optimal
  Primal bound      505757
  Dual bound        505757
[...]

PR #797, opened 2025-02-20
Construct lists before loop

christiansegercrantz:
Initializing the values in the loop and accessing them through the dictionary as done is very slow. This speeds up the operation from 11s/1000 constraints to ~0.5mil in sub 1 second. 

Related to issue https://github.com/coin-or/pulp/issues/793

(I am currently having env problems so unable to verify and test. Will do it later today/tommorrow)

### DIFF
--- a/pulp/apis/highs_api.py
+++ b/pulp/apis/highs_api.py
@@ -1,5 +1,5 @@
 # PuLP : Python LP Modeler
-# Version 2.4
+# Version 2.4.1
 from math import inf
 
 # Copyright (c) 2002-2005, Jean-Sebastien Roy (js@jeannot.org)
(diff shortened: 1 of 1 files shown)


christiansegercrantz:
I seem to be unable to request reviews, but would @pchtsp be able to review?

pchtsp (commented):
Thanks. It's not clear to me why this approach would be more efficient than the previous one. Do you have any idea?

It seems that `solution.row_value`, `solution.row_dual` are already lists. Why would you need to cast them into a list?

Other than that, I don't see any issue with the code itself.

christiansegercrantz:
The reason is that you are doing access operations every time. It's calling the row_value from the solution that is slow as far as I understand. As far as I know, it's not recommended have function calls in for statements. Like 
```python
for x in function():
    ...
```
is not great. Do 
```python
values = function()
for x in values:
    ...
```
instead. But this I am not 100% sure about. But the speed here is of multiple magnates. You/I can run a profiler on it if you like to show the diff.

pchtsp:
iterating over dict.values() is not inefficient. In fact, by casting it into a list, you're already incurring into some (small) additional time.

In any case, I take your word for it.

PR #800, opened 2025-02-21
Extra error checking and type hinting for FixedElasticSubProblem

MBradbury:


pulp/tests/test_pulp.py
```diff
@@ -421,7 +422,7 @@ def test_divide(self):
             z = LpVariable("z", 0)
             w = LpVariable("w", 0)
             prob += x + 4 * y + 9 * z, "obj"
-            prob += (2 * x + 2 * y).__div__(2.0) <= 5, "c1"
+            prob += ((2 * x + 2 * y) / 2.0) <= 5, "c1"
```
pchtsp, line 425: Can we add some kind of warning to users that the "/" cannot be used and for them to use __div__ from now on? (if that's the case, that is)

pchtsp (commented):
Thanks for also taking the time to add types everywhere. I have a minor question on a deprecation warning/ error on the use of "/". Thanks again.

pulp/tests/test_pulp.py
MBradbury: `__div__` is an old Python 2 operator and doesn't exist in Python 3 any more. The ones that do exist are [`__truediv__`](https://docs.python.org/3/library/operator.html#operator.__truediv__), [`__floordiv__`](https://docs.python.org/3/library/operator.html#operator.__floordiv__), [`__ifloordiv__`](https://docs.python.org/3/library/operator.html#operator.__ifloordiv__), [`__itruediv__`](https://docs.python.org/3/library/operator.html#operator.__itruediv__).

Again, this is probably another thing to look at implementing the three operators `__floordiv__`, `__ifloordiv__` and `__itruediv__`.

There are some bits of the code that still provide `__div__` and `__rdiv__`, which should be removed, as Python 3 won't use them.

pchtsp: oops I read it backwards I thought you were including the __div__ instead of the /. Understood on the needed changes for the __div__, we can add an issue on that. I'll merge this one.

PR #807, opened 2025-02-26
Type hints

MBradbury:
I have seen #755 and wanted to take a look at providing type hints also. I would appreciate a review of this, as there are some choices I made which need to be checked and others where a decision needs to be made.

Summary:
- Increases minimum version to Python 3.9, as this is when hinting `list[T]` was introduced (instead of using `List[T]`)
- Added Lpt* type aliases to avoid excessive Unions across the code
- Overloads for dict, dicts and matrix
- Most of the examples type check successfully
- Removed LpElement
- Needed to move away from dicts to dataclasses to serialise MPS. Ability to convert to dicts is provided by Python, but a separate package (dacite) has been needed to load from json into the dataclasses.
- Removed the ability for `LpAffineExpression to inherit from an OrderedDict, always inherit from dict.

Choices to make:
1. What type should be used to store coefficients? Currently a mixture of int and float is allowed.
2. dicts and matrix cannot be fully typed with Python's type hints. I have added overloads for 3 dimensional inputs, how many is appropriate?
3. What is `LpConstraintVar`? Is it a variable or is it a constraint? There are several pieces of code in `LpAffineExpression` that assume that all the items it contains are either a `LpVariable` or a `LpConstraintVar`. Several options:
    - Could `LpConstraintVar` inherit from `LpVariable`?
    - Manage dict of `LpConstraintVar` separately?
    - Does `LpConstraintVar` support operators in a `LpAffineExpression`?
4. There seems to be a number of problems with `FractionElasticSubProblem` as parameters seemed to be in the incorrect order for the `__init__`.

There is also a test failure to investigate (PULP_CBC_CMDTest.test_dual_variables_reduced_costs), but I am having trouble finding the root cause.

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -8,11 +8,8 @@ jobs:
     strategy:
       max-parallel: 21
       matrix:
-        python-version: ['3.8', '3.9', '3.10', '3.11', '3.12']
+        python-version: ['3.9', '3.10', '3.11', '3.12']
         os: [ubuntu-latest, macOS-latest, windows-latest]
-        exclude:
-          - os: macOS-latest
-            python-version: '3.8'
 
     steps:
       - uses: actions/checkout@v4
--- a/INSTALL
+++ b/INSTALL
@@ -4,7 +4,7 @@ Installation
 Note that to install PuLP you must first have a working python installation as
 described in `installing python`_.
 
-PuLP requires Python >= 2.7 or Python >= 3.4.
+PuLP requires Python >= 3.9.
 
 The latest version of PuLP can be freely obtained from github_.
 
@@ -15,7 +15,7 @@ By far the easiest way to install pulp is through the use of pip_.
 
 * In  windows (please make sure pip is on your path)::
 
-    c:\Python34\Scripts\> pip install pulp
+    c:\Python39\Scripts\> pip install pulp
 
 * In Linux::
 
--- a/README.rst
+++ b/README.rst
@@ -19,7 +19,7 @@ PuLP is part of the `COIN-OR project <https://www.coin-or.org/>`_.
 Installation
 ================
 
-PuLP requires Python 3.7 or newer.
+PuLP requires Python 3.9 or newer.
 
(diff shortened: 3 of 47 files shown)


christiansegercrantz:
Amazing work, thank you! I can also help with reviewing and changes if needed, although currently a bit busy.

Regarding python 3.8, it's already [end-of-lifecycle](https://devguide.python.org/versions/), so I don't see a reason not to move away from it, people shouldn't be using it anyway.

MBradbury:
For dacite, alternatively, pulp moves to pickle for disk serialisation which would eliminate the need for dacite, but then human-readability of the files would be lost.

The type hints do not prevent dicts / matrix being used with more dimensions, just means that IDEs won't provide accurate information on the returned type.

MBradbury:
I have been thinking about this and I think it would be best to split this up into different PRs:

1. The class hierarchy of LpElement, LpVariable, and LpConstraintVar
2. The loading and saving via dataclasses
3. Other general type hint work

pchtsp:
as reference @frrad has been doing some typing/annotations work here: https://github.com/coin-or/pulp/pull/819

MBradbury:
Yes we should, just two thing to get out of it before closing:
1. Need some kind of resolution to: https://github.com/coin-or/pulp/pull/825, it is not clear to me how to handle the failing tests due to the discussion in https://github.com/coin-or/pulp/pull/819 with @frrad 
2. Once that's addressed then I think the most valuable bit to extract is the typing of matrix and dicts https://github.com/coin-or/pulp/blob/7e5908413c32c0a61e6587c7a9849054df6f4efb/pulp/pulp.py#L285 as these were non-trivial to reason about

Other issues still need to be looked at, but may be better as individual issues:
1. LpVariable/LpConstraintVar both stored in LpAffineExpression, but do not implement the same interface
2. FractionElasticSubProblem looks like it has parameters in the incorrect order
3. What type should coefficients be stored as (with https://github.com/coin-or/pulp/issues/815 also adding Decimal to be considered)

MBradbury:
Closing this PR due to other type hinting work.

PR #812, opened 2025-03-19
Repackage

pchtsp:


pchtsp:
@frrad would you know why mypy is currently ignoring the pyproject.toml in github actions now that we're moving to using this file to build the package? I'm asking because it was you who added the mypy check in the beginning. Locally, it seems to work fine.

thanks!

frrad:
My guess would be this has something to do with `uv`. Have you tried something like `uv run mypy --config-file pyproject.toml ./` in case it's not pickup up the `pyprojec.toml` for some reason?

frrad:
Looks like you figured it out. What was the problem?

pchtsp:
to be 100% honest, I do not know. I tried so many things. I suspect that what worked was that I added the following to the `pyproject.toml`:
```
[[tool.mypy.overrides]]
module = [
  "pulp.*",
]
follow_imports = "skip"
```

PR #813, opened 2025-03-20
Avoid memory cost of saving initial lower and upper bounds if not fixing the value of a variable

MBradbury:
Continuing the theme of performance improvements, I have a model with a large number of variables (~4 million) which leads to high RAM usage when instantiating those variables.

`LpVariable` currently stores copies of the initial upper and lower bounds when instantiated. This is a cost that users need to pay even if they are not fixing/unfixing variables.

This change instead stores the original bounds when fixing instead. So the memory cost is paid only when fixing/unfixing variables.

### DIFF
--- a/pulp/pulp.py
+++ b/pulp/pulp.py
@@ -286,14 +286,14 @@ class LpVariable(LpElement):
         self, name, lowBound=None, upBound=None, cat=const.LpContinuous, e=None
     ):
         LpElement.__init__(self, name)
-        self._lowbound_original = self.lowBound = lowBound
-        self._upbound_original = self.upBound = upBound
+        self.lowBound = lowBound
+        self.upBound = upBound
         self.cat = cat
         self.varValue = None
         self.dj = None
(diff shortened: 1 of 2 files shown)


pchtsp:
The 
```python
        if not hasattr(self, "_lowbound_original") or not hasattr(
            self, "_upbound_original"
        ):
```
looks a bit hacky. Do you still see the increase in memory if you initialize both to None instead? Like that we at leas have the properties defined since the beginning? thanks!

MBradbury:
The cost seems to come from the attributes.

I'm happy to not go this route and close the PR. The RAM savings were relatively small, 50 MB for 3 million variables.

Ideally, `fix` would have returned the old bounds (e.g., as a context manager) that could have then been passed to `unfix`. This would mean the memory cost is only paid when needed. I only went this route to avoid breaking the API.

pchtsp:
If the savings were not very big, I prefer to close the PR. Thanks for this and all the help though!

Issue #814, opened 2025-03-21
Rounding error while reading variables from GLPK_CMD


unsatcore:
Integer (or all?) variables are being "rounded" to 6 most significant digits when using GLPK_CMD

Details for the issue
--------------------

#### What did you do?

Here is a minimal example

```python
from pulp import LpProblem, LpMaximize, LpVariable, getSolver, LpStatus
model = LpProblem('bad', LpMaximize)
Q = LpVariable('Q', cat='Integer', lowBound=0, upBound=123456789)
model += Q
model += Q >= 0
solver = getSolver('GLPK_CMD', timeLimit=10, keepFiles=True)
solver.options = ['-w', 'good.txt'] # could also add --exact here
model.solve(solver)
print(f'{model.status}, {LpStatus[model.status]}')
print(Q.value()) # prints 123457000
```

#### What did you expect to see?

The expected value for `Q` is `123456789` obviously

#### What did you see instead?

...however the actual output is `123457000` (note, this breaks the bounds and/or constraints)

Useful extra information
-------------------------

Program output
```
GLPSOL--GLPK LP/MIP Solver 5.0
Parameter(s) specified in the command line:
 --cpxlp bad-pulp.lp -o bad-pulp.sol --tmlim 10 -w good.txt
Reading problem data from 'bad-pulp.lp'...
1 row, 1 column, 1 non-zero
One variable is integer
10 lines were read
GLPK Integer Optimizer 5.0
1 row, 1 column, 1 non-zero
1 integer variable, none of which are binary
Preprocessing...
Objective value =   1.234567890e+08
INTEGER OPTIMAL SOLUTION FOUND BY MIP PREPROCESSOR
Time used:   0.0 secs
Memory used: 0.0 Mb (32525 bytes)
Writing MIP solution to 'bad-pulp.sol'...
Writing MIP solution to 'good.txt'...
11 lines were written
1, Optimal
123457000
```
In this particular instance the solution was found by MIP preprocessor.
However, in more realistic cases the solution could be found by actual MIP solver.
The point is the solver "knows" the solution precisely

`good.txt`:
```
c Problem:
c Rows:       1
c Columns:    1
c Non-zeros:  1
c Status:     INTEGER OPTIMAL
c Objective:  OBJ = 123456789 (MAXimum)
c
s mip 1 1 o 123456789
i 1 123456789
j 1 123456789
e o f
```

Some hint might be given by `bad-pulp.sol` output file:
```
Problem:
Rows:       1
[...]

PR #818, opened 2025-03-26
Fixes rounding error in GLPK_CMD api #814

unsatcore:
As described in #814 this PR implements reading column values from '--write' solution file instead of '--output'.

Alas it was not possible to get rid of '--output' file as `glpsol` provides no other means to determine column names.

Note 1, this could be a breaking change for some users

Note 2, there is still some precision limit as the original '.lp' file is being formatted with '.12g' specifier.
See e.g. [asCplexLpVariable](https://github.com/coin-or/pulp/blob/79f7146e3580baa3a4cf0e9e90b2129bd3622686/pulp/pulp.py#L575-L591) or [asCplexLpConstraint](https://github.com/coin-or/pulp/blob/79f7146e3580baa3a4cf0e9e90b2129bd3622686/pulp/pulp.py#L1098-L1116)

### DIFF
--- a/pulp/apis/glpk_api.py
+++ b/pulp/apis/glpk_api.py
@@ -118,9 +118,9 @@ class GLPK_CMD(LpSolver_CMD):
 
         status, values = self.readsol(tmpOut, tmpSol)
 
-        vars = dict([(var.name,var) for var in lp.variables()])
-        for name,value in values.items():
-            if name not in vars: # __dummy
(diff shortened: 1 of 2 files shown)


pchtsp:
thanks! Could you please run black on the code and push again?
https://coin-or.github.io/pulp/develop/contribute.html#applying-the-black-linter-formatter

PR #819, opened 2025-03-28
Type examples/

frrad:
Adds typing to files in `examples/`.

Also removed several more `pulp/` files from the mypy blacklist, adding `type: ignore`s where necessary.

### DIFF
--- a/examples/BeerDistributionProblem_resolve.py
+++ b/examples/BeerDistributionProblem_resolve.py
@@ -34,7 +34,7 @@ costs = [  # Bars
 ]
 
 # The cost data is made into a dictionary
-costs = makeDict([Warehouses, Bars], costs, 0)
+costs_dict = makeDict([Warehouses, Bars], costs, 0)
 
 # Creates the 'prob' variable to contain the problem data
 prob = LpProblem("Beer Distribution Problem", LpMinimize)
@@ -47,7 +47,7 @@ vars = LpVariable.dicts("Route", (Warehouses, Bars), 0, None, LpInteger)
 
 # The objective function is added to 'prob' first
 prob += (
-    lpSum([vars[w][b] * costs[w][b] for (w, b) in Routes]),
+    lpSum([vars[w][b] * costs_dict[w][b] for (w, b) in Routes]),
     "Sum_of_Transporting_Costs",
 )
 
--- a/examples/BeerDistributionProblem_resolve.py
+++ b/examples/BeerDistributionProblem_resolve.py
@@ -68,10 +68,10 @@ for b in Bars:
 
 # The problem data is written to an .lp file
 prob.writeLP("BeerDistributionProblem.lp")
-for demand in range(500, 601, 10):
+for demand_value in range(500, 601, 10):
     # reoptimise the problem by increasing demand at bar '1'
     # note the constant is stored as the LHS constant not the RHS of the constraint
-    bar_demand_constraint["1"].constant = -demand
+    bar_demand_constraint["1"].constant = -demand_value
     # or alternatively,
     # prob.constraints["Sum_of_Products_into_Bar_1"].constant = - demand
 
--- a/examples/CGcolumnwise.py
+++ b/examples/CGcolumnwise.py
@@ -22,7 +22,7 @@ class Pattern:
     lenOpts = ["5", "7", "9"]
     numPatterns = 0
 
-    def __init__(self, name: str, lengths: Optional[List[int]] = None) -> None:
+    def __init__(self, name: str, lengths: List[int]) -> None:
         self.name = name
         self.lengthsdict = dict(zip(self.lenOpts, lengths))
         Pattern.numPatterns += 1
@@ -89,7 +89,7 @@ def addPatterns(
         for j, k in zip(i, Pattern.lenOpts):
             lsum += j * int(k)
         if lsum > Pattern.totalRollLength:
-            raise ("Length Options too large for Roll")
+            raise Exception("Length Options too large for Roll")
 
         # The number of rolls of each length in each new pattern is printed
         print("P" + str(Pattern.numPatterns), "=", i)
@@ -99,15 +99,15 @@ def addPatterns(
 
     # The pattern variables are created
     pattVars = []
-    for i in Patterns:
+    for p in Patterns:
         pattVars += [
             LpVariable(
-                "Pattern " + i.name,
+                "Pattern " + p.name,
                 0,
                 None,
                 LpContinuous,
-                (i.cost - Pattern.trimValue * i.trim()) * obj
-                + lpSum([constraints[l] * i.lengthsdict[l] for l in Pattern.lenOpts]),
+                (p.cost - Pattern.trimValue * p.trim()) * obj
+                + lpSum([constraints[l] * p.lengthsdict[l] for l in Pattern.lenOpts]),
             )
         ]
 
--- a/examples/CGcolumnwise.py
+++ b/examples/CGcolumnwise.py
@@ -6,6 +6,8 @@ Authors: Antony Phillips,  Dr Stuart Mitchell  2008
 
 # Import PuLP modeler functions
 from pulp import *
+from pulp.pulp import LpConstraintVar, LpProblem
+from typing import Any, Dict, List, Optional, Tuple, Union
 
 
 class Pattern:
@@ -19,7 +21,7 @@ class Pattern:
     lenOpts = ["5", "7", "9"]
     numPatterns = 0
 
-    def __init__(self, name, lengths=None):
+    def __init__(self, name: str, lengths: Optional[List[int]]=None) -> None:
         self.name = name
         self.lengthsdict = dict(zip(self.lenOpts, lengths))
         Pattern.numPatterns += 1
@@ -27,13 +29,13 @@ class Pattern:
     def __str__(self):
         return self.name
 
-    def trim(self):
+    def trim(self) -> int:
         return Pattern.totalRollLength - sum(
             [int(i) * int(self.lengthsdict[i]) for i in self.lengthsdict]
         )
 
 
-def createMaster():
+def createMaster() -> Tuple[LpProblem, LpConstraintVar, Dict[str, LpConstraintVar]]:
     rollData = {  # Length Demand SalePrice
         "5": [150, 0.25],
         "7": [200, 0.33],
@@ -71,7 +73,7 @@ def createMaster():
     return prob, obj, constraints
 
 
-def addPatterns(obj, constraints, newPatterns):
+def addPatterns(obj: LpConstraintVar, constraints: Dict[str, LpConstraintVar], newPatterns: List[List[int]]) -> None:
     # A list called Patterns is created to contain all the Pattern class
     # objects created in this function call
     Patterns = []
@@ -105,7 +107,7 @@ def addPatterns(obj, constraints, newPatterns):
         ]
 
 
-def masterSolve(prob, relax=True):
+def masterSolve(prob: LpProblem, relax: bool=True) -> Union[Tuple[float, Dict[str, int]], Dict[str, float]]:
     # Unrelaxes the Integer Constraint
     if not relax:
         for v in prob.variables():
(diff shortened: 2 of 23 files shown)


examples/furniture.py
```diff
@@ -14,14 +14,14 @@
     [1, 2],  # Lathe
     [3, 1.5],  # Polisher
 ]
-activity = makeDict([Resources, Chairs], activity)
+activity_dict = makeDict([Resources, Chairs], activity)
```
pchtsp, line 17: I'm curious, is this some kind of naming convention to recognize the type objects? Why rename all dictionaries to have a `_dict` at the end?

examples/SpongeRollProblem4.py
```diff
@@ -80,14 +82,14 @@ class Pattern:
     totalRollLength = 20
     lenOpts = [5, 7, 9]
 
-    def __init__(self, name, lengths=None):
+    def __init__(self, name: str, lengths: List[int]) -> None:
```
pchtsp, line 85: shouldn't we keep the `=None` here?

examples/CGcolumnwise.py
```diff
@@ -19,21 +22,21 @@ class Pattern:
     lenOpts = ["5", "7", "9"]
     numPatterns = 0
 
-    def __init__(self, name, lengths=None):
```
pchtsp, line 22: also here, should we keep the `=None`?

examples/BeerDistributionProblem_resolve.py
```diff
@@ -68,10 +68,10 @@
 
 # The problem data is written to an .lp file
 prob.writeLP("BeerDistributionProblem.lp")
-for demand in range(500, 601, 10):
+for demand_value in range(500, 601, 10):
```
pchtsp, line 71: same as with `_dict`, is this some kind of naming convention based on the type of the object?

pchtsp (commented):
Thanks! I added minor questions about the renaming of some objects. I'm curious what tool renames objects based on their implied type. It sounds a bit weird.

christiansegercrantz:
> Thanks! I added minor questions about the renaming of some objects. I'm curious what tool renames objects based on their implied type. It sounds a bit weird.

This naming convention is in my opinion a bit outdated with modern typing. Instead of changing names to indicate type it would be better, as is being done in the linked PR, to indicate the type properly with typing.

MBradbury:
Few comments based on #807:
1. Why use `List` and not `list` now that 3.8 has been dropped?
2. I tried to check the type hints from the previous PR and different tools work differently. For example, I use pyright which is typically significant stricter than mypy and pyright still identifies some issues.
3. In #807, I also did a lot of typing work in examples/, its not likely that I will get round to fixing that PR, so might be worth taking a look and extracting what is needed.

pulp/pulp.py
```diff
@@ -144,28 +144,28 @@
 try:  # allow Python 2/3 compatibility
     maketrans = str.maketrans
 except AttributeError:
-    from string import maketrans
```
MBradbury, line 147: Why not just change to be: `from string import maketrans`, Python 2 is not supported

pulp/pulp.py
```diff
@@ -144,28 +144,28 @@
...
+    from string import maketrans  # type: ignore[attr-defined,no-redef]
 
 _DICT_TYPE = dict
 
 if sys.platform not in ["cli"]:
```
MBradbury, line 151: There is no `sys.platform` which is `cli` any more. I presume this was IronPython. Why not simplify and just remove `_DICT_TYPE` as per https://github.com/coin-or/pulp/pull/807/files#diff-b2a258edfe0fa294f032a978656a4082842558faa98b0950d9f68c50f546bbe2L151

pulp/pulp.py
```diff
@@ -1063,7 +1063,7 @@ def __init__(self, e=None, sense=const.LpConstraintEQ, name=None, rhs=None):
         """
         self.expr = e if isinstance(e, LpAffineExpression) else LpAffineExpression(e)
         self.name = name
-        self.constant: float = self.expr.constant
```
MBradbury, line 1066: This just hides a potential bug. Better to fix the annotations

https://github.com/coin-or/pulp/pull/807/files#diff-b2a258edfe0fa294f032a978656a4082842558faa98b0950d9f68c50f546bbe2R169

https://github.com/coin-or/pulp/pull/807/files#diff-b2a258edfe0fa294f032a978656a4082842558faa98b0950d9f68c50f546bbe2R1212

pulp/pulp.py
```diff
@@ -1218,7 +1218,7 @@ def __truediv__(self, other):
     def valid(self, eps: float = 0) -> bool:
         val = self.value()
         if self.sense == const.LpConstraintEQ:
-            return abs(val) <= eps
```
MBradbury, line 1221: Again, just hides the bug: https://github.com/coin-or/pulp/pull/807/files#diff-b2a258edfe0fa294f032a978656a4082842558faa98b0950d9f68c50f546bbe2R1370

MBradbury:
There are some easy wins in the type hinting space, but as I found in #807, there are some architectural issues with LpElement, LpVariable and LpConstraintVar that need to be fixed. Also FractionElasticSubProblem just seems fundamentally broken.

Then there is also the question of what type the coefficient should be? int, float, decimal, numpy int/float types, or what combination of these?

pulp/pulp.py
MBradbury: My links don't seem to have worked, there needs to be a check like:

```python
val = self.value()
if val is None:
    return False
```

pulp/pulp.py
MBradbury: In my PR, I added some type aliases, which I then used where appropriate.
```python
LptNumber: TypeAlias = int | float
LptItem: TypeAlias = Union[LptNumber, "LpVariable"]
LptExpr: TypeAlias = Union[LptItem, "LpAffineExpression"]
LptConstExpr: TypeAlias = Union[LptNumber, "LpAffineExpression"]
LptExprConstraint: TypeAlias = Union[LptExpr, "LpConstraint"]
LptVars: TypeAlias = "LpVariable"
```

examples/furniture.py
frrad: I changed the name of this variable because previously this line was changing the type of `activity` from a list to a dict. The `_dict` is a bit lazy admittedly. Very open to better name suggestions :)

examples/SpongeRollProblem4.py
frrad: I removed it because the `zip` below on L87 fails to typecheck if `lengths` is `None`. Since this code is presumably working I took that to mean that this is never used that way in practice.

examples/CGcolumnwise.py
frrad: as above

examples/BeerDistributionProblem_resolve.py
frrad: Similar to previously there is already a `demand` in scope and this was me fixing the collision.

examples/SpongeRollProblem5.py
```diff
@@ -21,23 +21,25 @@
...
-Patterns = []
+Patterns: List[Pattern] = []
 for i in patternslist:
     Patterns += [Pattern("P" + str(len(Patterns)), i)]
 
 # This loop will be repeated until morePatterns is set to False
 while morePatterns == True:
     # Solve the problem as a Relaxed LP
     duals = masterSolve(Patterns, rollData)
+    assert isinstance(duals, dict)
```
MBradbury, line 32: Surely, the return type is a tuple as per `masterSolve`?

examples/SpongeRollProblem5.py
```diff
@@ -21,23 +21,25 @@
 patternslist = [[4, 0, 0], [0, 2, 0], [0, 0, 2]]
 
 # The starting patterns are instantiated with the Pattern class
-Patterns = []
+Patterns: List[Pattern] = []
 for i in patternslist:
     Patterns += [Pattern("P" + str(len(Patterns)), i)]
 
 # This loop will be repeated until morePatterns is set to False
 while morePatterns == True:
```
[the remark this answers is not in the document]
frrad: I am trying to keep the scope of this PR as limited as possible. There's a ton of less-than-ideal code here. If I started fixing things I don't know where I would be able to stop.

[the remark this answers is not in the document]
MBradbury: Same problem I had in #807 :)

frrad:
👋 thanks for all the reviews.

To clear up some possible confusion - this PR is intended to be the minimum possible change to add types to the `examples/` directory. It is explicitly _not_ trying to do things like fix types in `pulp/` or improve the non-type code in `examples`. I think structuring things this way will lead to a change which is easy to review and a pareto improvement over the current situation. There will still definitely remain a lot of room to improve things afterwards, but by trying to be focused here I think we can hopefully get this one in without too much back and forth and make the next PR easier for whoever ends up making it.


Additional responses inline: 


> > Thanks! I added minor questions about the renaming of some objects. I'm curious what tool renames objects based on their implied type. It sounds a bit weird.
> 
> This naming convention is in my opinion a bit outdated with modern typing. Instead of changing names to indicate type it would be better, as is being done in the linked PR, to indicate the type properly with typing.

Sorry for the confusion caused by my clunky variable names. They are not meant to convey the type, just fix the variable type change issues. These variables all actually typecheck fine without adding explicit hints.

> re: changes to `pulp`

These are all great questions. The intention of this PR is not to fix any type problems in `pulp/`. I just added whatever `type: ignore` were necessary to allow me to remove these files from the mypy blacklist. I definitely agree that a lot more work is needed to actually fix the issues there.

MBradbury:
> > re: changes to `pulp`
> 
> These are all great questions. The intention of this PR is not to fix any type problems in `pulp/`. I just added whatever `type: ignore` were necessary to allow me to remove these files from the mypy blacklist. I definitely agree that a lot more work is needed to actually fix the issues there.

Would it be better to **not** add the `# type: ignore`s then? By adding them potential bugs are covered up. IMO, better to let the tests fail and then fix the bug than to hide the issue.

frrad:
In general unit tests / CI work best if you keep them in a state where you always expect them to pass. You want their binary pass / fail state to be a useful signal of whether or not a new change introduces new bugs. If you add tests which you expect to fail it can confuse things quite a bit.

The type ignore statements are a useful crutch to be able to make incremental improvements without having to fix everything at once. With their help we can do 

1. The current state mypy doesn't even consider these files and CI gives us no information about what changes do to typing in them
2. After we add ignores readers of the code will at least be aware of existing type issues. 
3. Any future PRs which touch these files will fail CI if they introduce new type errors.
4. New PRs can be made which fix the existing type issues and remove the ignore comments. Reviewers will be able to be confident in the correctness of these PRs because CI status will tell them whether or not mypy passes

christiansegercrantz:
> > > Thanks! I added minor questions about the renaming of some objects. I'm curious what tool renames objects based on their implied type. It sounds a bit weird.
> > 
> > 
> > This naming convention is in my opinion a bit outdated with modern typing. Instead of changing names to indicate type it would be better, as is being done in the linked PR, to indicate the type properly with typing.
> 
> Sorry for the confusion caused by my clunky variable names. They are not meant to convey the type, just fix the variable type change issues. These variables all actually typecheck fine without adding explicit hints.

Makes sense, didn't read the code in larger context. Good change then, at some point those should probably have completely different names to convey their meaning better. But a good first step.

frrad:
merged master to pick up https://github.com/coin-or/pulp/pull/820 - looks like I have some work to do

frrad:
That wasn't too bad at least 😅 

@pchtsp I think I have addressed all your comments. Please let me know if there is anything more I can answer or change.

There are some references to these files in the documentation by line number like this: https://github.com/coin-or/pulp/blob/f626d34f9b345ffc8bf405f9711d160fe20a4f1d/doc/source/CaseStudies/a_blending_problem.rst?plain=1#L97-L98

which may need to be updated if things have shifted around. I'm happy to have a look at this as well as part of this PR. I want to wait until the actual code changes look good first though in case the lines end up moving around more.

pchtsp:
thanks @frrad . I don't mind if we do not have a perfect state now. Anything that gets as closer is good. 
@christiansegercrantz , @MBradbury thanks for your help, if you can make any improvement on this on a separate PR (e.g., through https://github.com/coin-or/pulp/pull/807/) I can review it and merge it. Again, if we do it in pieces, we will probably finish sooner.

Thanks again

PR #821, opened 2025-03-31
get available mip solution if status is not optimal

riis-hhd:


riis-hhd:
Sure. To repro, solve a mip problem with timelimit. The  feasible solution might be available after timeout. A python test file is attached here.
[repro.txt](https://github.com/user-attachments/files/19543152/repro.txt)

image

PR #827, opened 2025-04-05
fix set partition problem case study docs

frrad:
Currently this page https://coin-or.github.io/pulp/CaseStudies/a_set_partitioning_problem.html from the docs is messed up as a result of https://github.com/coin-or/pulp/pull/819 moving some line numbers around.

This PR fixes the docs and moves to a new style which uses special comments to avoid this happening again in the future. The potential drawback is that the bit at the end where it includes the full file content will now also include these special comments and look a bit funky. I personally think it's worth it but curious on your take.

I suspect there are other docs which are damaged in this same way. Happy to fix but wanted to get your take on the new approach before I do the rest.


pchtsp:
Thanks. I think it's ok, go ahead with the rest of examples.

frrad:
I converted one further example and verified that https://github.com/coin-or/pulp/pull/819 does not move lines around in the remaining files which use `literalInclude` by line number (`a_two_stage_production_planning_problem.rst`, `a_transportation_problem.rst`, `a_sudoku_problem.rst`).

PR #830, opened 2025-04-09
remove more of the mypy ignore list

frrad:
This removes the last of the "real" exceptions from `pyproject.toml`, adding `type: ignore` as needed.

### DIFF
--- a/pulp/mps_lp.py
+++ b/pulp/mps_lp.py
@@ -158,9 +158,9 @@ def readMPS(path: str, sense: int, dropConsNames: bool = False) -> MPS:
 
     with open(path) as reader:
         for line in reader:
-            line = re.split(" |\t", line)
-            line = [x.strip() for x in line]
-            line = list(filter(None, line))
+            line = re.split(" |\t", line)  # type: ignore[assignment]
+            line = [x.strip() for x in line]  # type: ignore[assignment]
+            line = list(filter(None, line))  # type: ignore[assignment]
 
             if line[0] == "ENDATA":
                 break
@@ -229,9 +229,9 @@ def readMPS(path: str, sense: int, dropConsNames: bool = False) -> MPS:
                     raise const.PulpError(
                         "Other RHS name was given even though name was set after RHS tag."
                     )
-                readMPSSetRhs(line, constraints)
+                readMPSSetRhs(line, constraints)  # type: ignore[arg-type]
             elif mode == CORE_FILE_RHS_MODE_NO_NAME:
-                readMPSSetRhs(line, constraints)
+                readMPSSetRhs(line, constraints)  # type: ignore[arg-type]
                 if line[0] not in rhs_names:
                     rhs_names.append(line[0])
             elif mode == CORE_FILE_BOUNDS_MODE_NAME_GIVEN:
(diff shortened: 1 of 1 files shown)


pulp/apis/gurobi_api.py
```diff
@@ -163,10 +163,9 @@ def __init__(
 
             # set the output of gurobi
             if not self.msg:
```
frrad, line 165: this is unrelated, just saw it as I was reading the code.

pulp/tests/test_gurobipy_env.py
```diff
@@ -14,6 +14,15 @@ def check_dummy_env():
         pass
 
 
+def is_single_use_license() -> bool:
```
frrad, line 17: the `SkipTest` decorators below failed to type check. I was deciding between this and just unilateral skip with `@unittest.skip`. Happy to go the other way if you want.

pulp/tests/test_pulp.py
```diff
@@ -1574,7 +1576,7 @@ def test_constraint_add(self):
             self.assertEqual(str(c1_variable), str(2 * x + y <= 5))
             self.assertEqual(repr(c1_variable), repr(2 * x + y <= 5))
 
-            expr: LpAffineExpression = x + 1
```
frrad, line 1577: mypy complains about these type hints since these variables are already defined.

pulp/tests/test_gurobipy_env.py
pchtsp: I don't really understand the difference between `unittest.SkipTest` and `unittest.skip`. Maybe @torressa knows ?

pchtsp: Is there a way to check for the single use license inside the `is_single_use_license` function and return true/false?

torressa: With `gurobipy` one could do:

```python
def is_single_use_license() -> bool:
    try:
        env1 = gp.Env()
        env2 = gp.Env()
        env2.dispose()
        env1.dispose()
    except gp.GurobiError as ge:
        if ge.errno == gp.GRB.Error.NO_LICENSE:
            print("single use license")
            print(f"Error code {ge.errno}: {ge}")
            return True
    return False
```

This is enough to trigger an error in a single-use case.

frrad:
I have put in the implementation supplied by @torressa. Thanks!

On my machine these tests are still skipped. If you have a single-use license environment to play with it might be good to verify that they are run and pass. 

```
pytest -rs pulp/tests/test_gurobipy_env.py
================================================== test session starts ===================================================
platform linux -- Python 3.10.14, pytest-8.3.5, pluggy-1.5.0
rootdir: /home/frederick/Projects/pulp
configfile: pyproject.toml
plugins: darkgraylib-1.2.1
collected 7 items                                                                                                        

pulp/tests/test_gurobipy_env.py s.ss...                                                                            [100%]

================================================ short test summary info =================================================
SKIPPED [1] pulp/tests/test_gurobipy_env.py:83: this test is only expected to pass with a single-use license
SKIPPED [1] pulp/tests/test_gurobipy_env.py:56: this test is only expected to pass with a single-use license
SKIPPED [1] pulp/tests/test_gurobipy_env.py:123: this test is only expected to pass with a single-use license
============================================== 4 passed, 3 skipped in 0.23s ==============================================
```

pchtsp:
it seems to still fail with the current changes, @frrad could you take a look? thanks!

frrad:
It was working! I wonder what changed 🤔 will take a look.

frrad:
Fixed! Looks like some minor conflict with https://github.com/coin-or/pulp/pull/831/

pchtsp (approved):

PR #833, opened 2025-04-23
Fix Xpress solver deprecation warnings for API changes in Xpress 9.5

pepeleduin:
Xpress 9.5 is using [problem.getSolution](https://www.fico.com/fico-xpress-optimization/docs/latest/solver/optimizer/python/HTML/problem.getSolution.html) instead of problem.getmipsol. Also, ``sense`` attribute in xpress.constraint has changed to [``type`` attribute](https://www.fico.com/fico-xpress-optimization/docs/latest/solver/optimizer/python/HTML/xpress.constraint.html).

### DIFF
--- a/pulp/apis/xpress_api.py
+++ b/pulp/apis/xpress_api.py
@@ -456,8 +456,8 @@ class XPRESS_PY(LpSolver):
                     2: constants.LpStatusUndefined,  # XPRS_MIP_LP_OPTIMAL
                     3: constants.LpStatusUndefined,  # XPRS_MIP_NO_SOL_FOUND
                     4: constants.LpStatusUndefined,  # XPRS_MIP_SOLUTION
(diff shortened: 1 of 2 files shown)


pchtsp:
the unittests seem to be failing for the XPRESS API. Have you run the tests locally? In that case, can you provide backwards  compatible code that checks the version of XPRESS being used so that it doesn't break with older versions?

pchtsp:
@pepeleduin can you fix the issues I mentioned? Otherwise I'll have to close the PR. Thanks again,

pepeleduin:
Hello @pchtsp . In this case I added more robust snippets to handle the Xpress version. However, there are two deprecated warnings that still I cannot solve, for test_integer_infeasible and test_integer_infeasible_2.

PR #836, opened 2025-05-05
Update time limit flag from -tl to -limit [-{time}s]

pepeleduin:
The previous flag was not working for CHOCO v5.0.0


pchtsp:
Thanks. I'm curious on what is use case of using CHOCO with PuLP instead of other free MIP solvers.

pepeleduin:
> Thanks. I'm curious on what is use case of using CHOCO with PuLP instead of other free MIP solvers.

In my case, I'm making benchmarking haha nothing special for CHOCO, unless its performance (time and resources) becomes remarkable.

PR #837, opened 2025-05-07
CuOpt Solver Integration

Iroy30:


pulp/apis/cuopt_api.py
```diff
@@ -0,0 +1,337 @@
...
+    clock,
+    ctypesArrayFill,
+    sparse,
+)
+
+# COPT string convention
+if sys.version_info >= (3, 0):
+    coptstr = lambda x: bytes(x, "utf-8")
+else:
+    coptstr = lambda x: x
```
pchtsp, line 36: PuLP requires python >= 3.9 so no need to add this conditional

pulp/apis/cuopt_api.py
```diff
@@ -0,0 +1,337 @@
...
+
+def process_constraint(args):
+    local_data = []
+    local_vars = []
+    local_sense = []
+    local_rhs = []
+    for constraints in args:
+        name, constraint = constraints
+        local_data.extend(list(constraint.values()))
+        local_vars.extend(list(constraint.keys()))
```
pchtsp, line 54: I think the order of keys and values dictionaries in python is preserved so this code *should* run fine.
Having said that, I think I prefer the following code, that takes values and keys in one go, in the remote case the order is not preserved for whatever reason.
```python
data, vars = zip(*a.items())
local_data.extend(data)
local_vars.extend(vars)
```

pulp/apis/cuopt_api.py
```diff
@@ -0,0 +1,337 @@
...
+        local_sense.append(sense_conv[constraint.sense])
+        local_rhs.append(-constraint.constant)
+    return local_data, local_vars, local_sense, local_rhs
+
+
+def process_variables(var):
+    #obj_coeff = lp.objective.get(var, 0.0))
```
pchtsp, line 61: delete this line if it's not needed

pulp/apis/cuopt_api.py
```diff
@@ -0,0 +1,337 @@
+import ctypes
+import os
+import subprocess
+import sys
+import warnings
+from uuid import uuid4
+import numpy as np
```
pchtsp, line 7: numpy is not part of PuLP dependencies. So this import will fail if cuopt is not installed.

pulp/apis/cuopt_api.py
```diff
@@ -0,0 +1,337 @@
...
+
+        def __init__(
+            self,
+            mip=True,
+            msg=True,
+            timeLimit=None,
+            gapRel=None,
+            warmStart=False,
+            logPath=None,
+            **solverParams,
```
pchtsp, line 108: solverParams is not defined in the docstring

pulp/apis/cuopt_api.py
```diff
@@ -0,0 +1,337 @@
...
+            self.solver_params = solverParams
+
+            ## TODO: Disable logging if self.msg = False
+            for key, value in solverParams.items():
+                if key == "optimality_tolerance":
+                    self.settings.set_optimality_tolerance(value)
+
+
+        def findSolutionValues(self, lp):
+            model = lp.solverModel
```
pchtsp, line 144: `model` does not seem to be used inside this method.

pulp/apis/cuopt_api.py
```diff
@@ -0,0 +1,337 @@
...
+            return True
+
+        def callSolver(self, lp, callback=None):
+            """Solves the problem with CUOPT"""
+            self.solveTime = -clock()
+            # TODO: Add callback
+            log_file = self.optionsDict.get("logPath") or ""
+
+            self.settings.set_infeasibility_detection(True)
+            self.solution = self.solver.Solve(lp.solverModel, self.settings, log_file)
```
pchtsp, line 210: most other solvers store the solution inside `lp.solverModel`, I see this is not the case for cuopt by default.

Instead of creating an attribute  `self.solution` to store it in the solver, I think it's cleaner if callSolver returns the content of `self.solver.Solve` and that gets passed to `findSolutionValues` as an additional argument named `solution`. 

Another option is to attach the `solution` attribute to `lp.solverModel` and then you get everything from the object `lp`.

pulp/apis/cuopt_api.py
```diff
@@ -0,0 +1,337 @@
...
+            lp.solverModel.set_variable_names(np.array(var_name))
+
+            var_warmstart = []
+            if self.optionsDict.get("warmStart", False):
+                for var in lp._variables:
+                    if var.varValue is not None:
+                        var_warmstart.append(var.varValue)
+                    else:
+                        var_warmstart = None
+                        break
```
pchtsp, line 267: does this work? I'm not sure this does what it's supposed to do. If one variable has a None value then the list is incomplete.

pulp/apis/cuopt_api.py
```diff
@@ -0,0 +1,337 @@
...
+
+            #if var_warmstart:
+            #    print("Setting variable warmstart: ", var_warmstart)
+            #    lp.solverModel.set_initial_primal_solution(np.array(var_warmstart))
```
pchtsp, line 271: to delete?

pulp/apis/cuopt_api.py
```diff
@@ -0,0 +1,337 @@
...
+                matrix_data.extend(list(constraint.values()))
+                matrix_indices.extend([var_dict[v.name] for v in constraint.keys()])
+                matrix_indptr.append(len(matrix_data))
+                try:
+                    c_sense = sense_conv[constraint.sense]
+                except:
+                    raise PulpSolverError("Detected an invalid constraint type")
+                rhs.append(-constraint.constant)
+                sense.append(c_sense)
+                #constraint.solverConstraint = {name: {"bound": -constraint.constant, "sense": c_sense, "coefficients": row_coeffs}}
```
pchtsp, line 287: to delete?

pulp/apis/cuopt_api.py
```diff
@@ -0,0 +1,337 @@
...
+                msg=msg,
+                timeLimit=timeLimit,
+                gapRel=gapRel,
+                logPath=logPath,
+                warmStart=warmStart,
+            )
+
+            from cuopt.linear_programming import data_model, solver_settings, solver
+            self.model = data_model.DataModel()
+            self.settings = solver_settings.SolverSettings()
```
pchtsp, line 131: why is it needed to define the settings during the initialization? I think all the settings management can be done inside the `callSolver` method without the need to creating and maintaining a `settings` attribute.

pulp/apis/cuopt_api.py
```diff
@@ -0,0 +1,337 @@
...
+            """True if the solver is available"""
+            return True
+
+        def callSolver(self, lp, callback=None):
+            """Solves the problem with CUOPT"""
+            self.solveTime = -clock()
+            # TODO: Add callback
+            log_file = self.optionsDict.get("logPath") or ""
+
+            self.settings.set_infeasibility_detection(True)
```
pchtsp, line 209: here, you could have everything related to settings:

```python
settings = solver_settings.SolverSettings()
settings.set_infeasibility_detection(True)
if self.timeLimit:
    settings.set_time_limit(self.timeLimit)
for key, value in solverParams.items():
    if key == "optimality_tolerance":
        settings.set_optimality_tolerance(value)
gapRel = self.optionsDict.get("gapRel")
if gapRel:
    self.settings.set_relative_gap_tolerance(gapRel)
```

Alternatively, you can attach the `settings` attribute to the `solverModel` object (as with the `solution` attribute). And edit it in `buildSolverModel`

pulp/tests/test_pulp.py
```diff
@@ -98,7 +98,11 @@ class PuLPTest(unittest.TestCase):
         solveInst = None
 
         def setUp(self):
-            self.solver = self.solveInst(msg=False)
+            if self.solveInst == CUOPT:
+                # cuOpt requires a user provided time limit for MIP problems
+                self.solver = self.solveInst(msg=False, timeLimit=120)
```
pchtsp, line 103: if the argument is mandatory for MIP problems, then a check needs to be included somewhere in the API to check if `timeLimit` is provided by the user before trying to solve.

pchtsp (commented):
thanks for the contribution. I've added some minor style requests, some required changes (e.g., numpy), some doubts (e.g., on warmstart), and other design preferences.

pulp/tests/test_pulp.py
amas008: It would seem somewhat unusual for a time limit to be required. Gurobi, for example, says:
https://docs.gurobi.com/projects/optimizer/en/current/reference/parameters.html#parametertimelimit
**TimeLimit**
Type: double
Default value: **Infinity**
It would seem to be better if CuOpt could follow this standard approach.

pulp/apis/cuopt_api.py
Iroy30: Removed in #839

pulp/apis/cuopt_api.py
Iroy30: Removed in #839

pulp/apis/cuopt_api.py
Iroy30: removed in #839

pulp/apis/cuopt_api.py
Iroy30: Moved to import while importing cuopt in #839

pulp/apis/cuopt_api.py
Iroy30: Added docstring in #839

pulp/apis/cuopt_api.py
Iroy30: Removed in #839

pulp/apis/cuopt_api.py
Iroy30: Added solution as return object from callSolver. Refer #839

pulp/apis/cuopt_api.py
Iroy30: Removed for now. Might add back in future.

pulp/apis/cuopt_api.py
Iroy30: Deleted in #839

pulp/apis/cuopt_api.py
Iroy30: Deleted in #839

pulp/apis/cuopt_api.py
Iroy30: Moved solver_settings to CallSolver. Refer #839

pulp/apis/cuopt_api.py
Iroy30: Updated in #839

pulp/tests/test_pulp.py
Iroy30: New cuOpt version now doesn't need mandatory timelimit for MIP. Removed in #839

Iroy30:
Closing in favor of PR #839 from Nvidia-pulp

PR #839, opened 2025-05-19
CuOpt Solver Integration

Iroy30:


Iroy30:
Incorporated review comments from #837

pchtsp:
Thanks. I'm not sure why you closed the previous PR. As long as you push the code to the branch, you can leave the PR open and it will update with the new changes. This makes easier to follow the changes/ suggestions and comments.

For example, there seems to be an error with the types when checking them with `mypy`, as I mentioned in the previous PR. I've added instructions to run mypy locally in the contributing site: https://coin-or.github.io/pulp/develop/contribute.html#checking-types-with-mypy

Also, you seem to not have signed the CLA in this PR.

bbozkaya:
Hi, we have re-submitted this PR via the NVidia fork of the Pulp repo using NV org account, as discussed with COIN-OR and agreed upon. All the reviews requested for the original PR are addressed.
https://github.com/coin-or/pulp/pull/839

pulp/apis/cuopt_api.py
```diff
@@ -0,0 +1,286 @@
...
+                self,
+                mip=mip,
+                msg=msg,
+                timeLimit=timeLimit,
+                gapRel=gapRel,
+                logPath=logPath,
+                warmStart=warmStart,
+            )
+
+            from cuopt.linear_programming import data_model
```
pchtsp, line 94: change for:
```
            from cuopt.linear_programming import data_model  # type: ignore[import-not-found, import-untyped, unused-ignore]

```

pulp/apis/cuopt_api.py
```diff
@@ -0,0 +1,286 @@
...
+class CUOPT(LpSolver):
+    """
+    The CUOPT Optimizer via its python interface
+    """
+
+    name = "CUOPT"
+
+    try:
+        global cuopt
+        import cuopt  # type: ignore[import-not-found]
```
pchtsp, line 48: change for:
```
import cuopt  # type: ignore[import-not-found, import-untyped, unused-ignore]
```

pchtsp (commented):
as I mentioned, running `mypy ./` is still failing. I've added the lines that need to be changed in the code.

pulp/apis/cuopt_api.py
Iroy30: Updated, thanks!

Iroy30:
@pchtsp Currently I am running all the pulp tests in test_pulp.py with CUOPT as the solver instance (Updated the test in the PR). I can paste the results as a comment here since the CI cannot run cuopt. Is there anything else we need to verify wrt testing?

Iroy30:
Also mypy has started failing again. I can fix the type issue in test_pulp.py , however, I cannot reproduce the numpy import issue in apis/cuopt_api.py
Could you confirm it would be the right place to do a numpy import ?

pchtsp:
hello, thanks for the reminder to look into this. Two things:

1. I fixed the typing issue in the master branch, you should just need to pull the master again.
2. for numpy I'd recommend just doing the same as for the other libraries: `import numpy as np  # type: ignore[import-not-found, import-untyped, unused-ignore]` and hope it doesn't complain.

thanks and let's hope this is the last change :  )

PR #850, opened 2025-07-21
Add callback and parameter configuration support to CPLEX_PY solver

G-Carneiro:
This pull request includes several updates and improvements related to the CPLEX solver:

- Updated the `CPLEX_PYTest` suite to use `CPLEX_PY` instead of `CPLEX_CMD`.
- Extended the solver API to accept custom CPLEX solver parameters via the `solverParams` argument, offering more control. Added utility methods to manage these parameters, along with unit tests to ensure proper behavior.
- Introduced support for registering CPLEX callbacks by adding an optional `callback` parameter. Included a test case to validate this functionality.

### DIFF
--- a/pulp/apis/cplex_api.py
+++ b/pulp/apis/cplex_api.py
@@ -292,7 +292,7 @@ class CPLEX_PY(LpSolver):
 
             Parameters should use dot notation as specified in the CPLEX documentation.
             The 'parameters.' prefix is optional. For example:
-                  
+
                   * parameters.advance (or advance)
                   * parameters.barrier.algorithm (or barrier.algorithm)
                   * parameters.mip.strategy.probe (or mip.strategy.probe)
@@ -513,8 +513,11 @@ class CPLEX_PY(LpSolver):
             """
             self.solverModel.parameters.timelimit.set(timeLimit)
 
-        def callSolver(self, isMIP,
-                       callback: Optional[Iterable[type[cplex.callbacks.Callback]]] = None):
+        def callSolver(
+            self,
+            isMIP,
+            callback: Optional[Iterable[type[cplex.callbacks.Callback]]] = None,
+        ):
             """
             Solves the problem with cplex
 
--- a/pulp/apis/cplex_api.py
+++ b/pulp/apis/cplex_api.py
@@ -448,14 +448,26 @@ class CPLEX_PY(LpSolver):
                 self.set_param(param, value)
 
         def set_param(self, name: str, value):
+            """
+            Sets a parameter value using its name.
+            """
             param = self.search_param(name=name)
             param.set(value)
 
         def get_param(self, name: str):
+            """
+            Returns the value of a named parameter by searching within the instance's parameters.
+            """
             param = self.search_param(name=name)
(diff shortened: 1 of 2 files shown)


pulp/apis/cplex_api.py
```diff
@@ -297,22 +308,25 @@ def __init__(
...
                 threads=threads,
             )
+            self.solverParams = solverParams
 
         def available(self):
             """True if the solver is available"""
             return True
 
-        def actualSolve(self, lp, callback=None):  # type: ignore[misc]
+        def actualSolve(self, lp, callback: Optional[Iterable[type[cplex.callbacks.Callback]]] = None):  # type: ignore[misc]
```
pchtsp, line 317: does this mean you can take out the ignore[misc] comment? 
Do you have the documentation that refers to the type `Iterable[type[cplex.callbacks.Callback]]`?

pulp/apis/cplex_api.py
```diff
@@ -430,6 +444,47 @@ def cplex_var_types(var):
...
+
+        def search_param(self, name: str):
+            """
+            Searches for a solver model parameter by its name and returns the corresponding attribute.
+
+            The method takes a parameter name string, processes it to remove the "parameters." prefix
+            and splits it by periods to traverse the attribute hierarchy of the solver model's parameters.
+            """
+            name = name.replace("parameters.", "")
+            param = self.solverModel.parameters
```
pchtsp, line 472: `self.solverModel` may not exist if `self.buildSolverModel` has not been called yet. Maybe we need to initialized it with None at __init__ and then check if it's none before using it

pulp/apis/cplex_api.py
```diff
@@ -297,22 +308,25 @@ def __init__(
                 logPath=logPath,
                 threads=threads,
             )
+            self.solverParams = solverParams
```
pchtsp, line 311: please use `self.solver_params` so it's similar to the gurobi API

pulp/apis/cplex_api.py
```diff
@@ -458,9 +513,21 @@ def setTimeLimit(self, timeLimit=0.0):
             """
             self.solverModel.parameters.timelimit.set(timeLimit)
 
-        def callSolver(self, isMIP):
-            """Solves the problem with cplex"""
+        def callSolver(
+            self,
+            isMIP,
```
pchtsp, line 518: I think `isMIP` should be type `bool`

pchtsp:
also, can you merge back the master branch? I just fixed an issue with one of the solver dependencies and it should allow running the tests again

pulp/apis/cplex_api.py
G-Carneiro: I think ignore[misc] is still needed, since the original definition in LPSolver doesn't have `callback`.
Docs:
- https://docs.python.org/3.12/library/collections.abc.html#collections.abc.Iterable
- https://docs.python.org/3.12/library/typing.html#the-type-of-class-objects
- https://www.ibm.com/docs/en/icos/22.1.1?topic=optimizers-cplex-python-api-reference-manual

PR #869, opened 2025-08-26
Check if the solution variables content is not None

pepeleduin:
When running pytest there is a TypeError: unsupported operand type(s) for -: 'NoneType' and 'int'. This is because there is no checking of the content of variables before doing operations with integers.


pchtsp:
thanks, can you sign the license?

pepeleduin:
Done!

pchtsp:
I think you should raise a PulpError if the variable has the value None, because it should fail. With your current modification the test would pass when `v.varValue` is None.

pchtsp:

PR #877, opened 2025-10-16
Validate solver solution count before processing

### LINKED ISSUE #876: SCIP_PY reporting inexistent solution
Details for the issue
--------------------

##### ... HARD PROBLEM THAT TAKES MANY SECONDS TO SOLVE

prob.solve(lp.SCIP_PY(**{
    "msg": True,
    'options': [
        'limits/time', 4
    ]
}))
print('Status', prob.status)
print("Translated Status:", LpStatus[prob.status])
###### Translated Status: Optimal

Meanwhile internally:

status = scip_to_pulp_status[solutionStatus]
print(f'{solutionStatus=} {status=}')

##### Reports: solutionStatus='timelimit' status=0

num_solutions = lp.solverModel.getNSols()
print(f'=====> {num_solutions=}')

If number of solutions is 0 and the solver returns timelimit reached pulp should not report 'optimal' and allow access to model (that does not meet the constraints)

You can access a model that does not obey the constraints

Useful extra information
-------------------------

- [X] Windows: ( _version:11)

- [ X] 3.12

#### I installed PuLP via:

- [X] pypi (python -m pip install pulp)

#### Did you also

- [x] I'm submitting a tentative PR that should fix this issue

josalhor:
Add validation for the number of solutions returned by the solver.

Should fix #876

This PR has not been tested (and I'm actually a little bit unsure why there is the try catch in the first place).
But the solution probably goes in this direction.

### DIFF
--- a/pulp/apis/scip_api.py
+++ b/pulp/apis/scip_api.py
@@ -565,6 +565,8 @@ class SCIP_PY(LpSolver):
             if solutionStatus in possible_solution_found_statuses:
                 try:  # Feasible solution found
                     solution = lp.solverModel.getBestSol()
+                    if lp.solverModel.getNSols() <= 0:
+                        raise ValueError('Returned solution by solver cannot be trusted')
                     for variable in lp._variables:
                         variable.varValue = solution[variable.solverVar]
                     for constraint in lp.constraints.values():


pchtsp:
thanks, can you provide a unit test with a toy example that currently fails and should pass?

PR #890, opened 2025-11-30
docs: add docstrings to .matrix() and .dict() methods

urbanophile:
- Add docstrings for .matrix() and .dict() methods
- Add more detail to docstring for .dicts() method
- Add basic usage examples
- Add some type annotations

### DIFF
--- a/pulp/pulp.py
+++ b/pulp/pulp.py
@@ -396,13 +396,42 @@ class LpVariable(LpElement):
     @classmethod
     def matrix(
         cls,
-        name,
-        indices=None,
-        lowBound=None,
-        upBound=None,
-        cat=const.LpContinuous,
-        indexStart=[],
-    ):
+        name: str,
+        indices: tuple = None,
+        lowBound: Optional[float] = None,
+        upBound: Optional[float] = None,
+        cat: str = const.LpContinuous,
+        indexStart: Optional[list] = [],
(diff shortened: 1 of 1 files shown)


pulp/pulp.py
```diff
@@ -396,13 +396,44 @@ def add_expression(self, e):
...
         cls,
-        name,
-        indices=None,
-        lowBound=None,
-        upBound=None,
-        cat=const.LpContinuous,
-        indexStart=[],
-    ):
+        name: str,
+        indices: Iterable,
```
pchtsp, line 400: here `indices` should be optional to match the original definition. it's also the case for the `dicts` classmethod

pulp/pulp.py
```diff
@@ -443,20 +474,33 @@ def dicts(
...
+            indexStart = []
         if not isinstance(indices, tuple):
             indices = (indices,)
         if "%" not in name:
             name += "_%s" * len(indices)
 
         index = indices[0]
         indices = indices[1:]
-        d = {}
+        d: dict[Any, Any] = {}
```
pchtsp, line 503: I guess here you can do something like `dict[Any, Union[dict, LpVariable]]`

pchtsp (commented):
Thanks, I left some minor comments on optional and union types

pchtsp:

PR #901, opened 2026-04-16
Handle cuOpt UnboundedOrInfeasible status (11) in CUOPT adapter and tests

ramakrishnap-nv:
## Summary

cuOpt added a new termination status (value 11, \`UnboundedOrInfeasible\`) that its presolver returns when infeasibility and unboundedness cannot be disambiguated. PuLP has no combined status, so the adapter maps it to \`LpStatusUndefined\` (mirroring how \`GLPK_CMD\` reports the same case). Status 10 (\`WorkLimit\`) is added explicitly at the same time.

Two test branches are updated so \`CUOPT\` is accepted in the tolerant paths that \`GLPK_CMD\` / \`GUROBI\` / \`MOSEK\` already use:

- \`test_integer_infeasible_2\`: accept \`Infeasible\` or \`Undefined\` for \`CUOPT\`
- \`test_unbounded\`: accept \`Unbounded\` or \`Undefined\` for \`CUOPT\`

Tracked upstream in NVIDIA/cuopt#1114.

### DIFF
--- a/pulp/apis/cuopt_api.py
+++ b/pulp/apis/cuopt_api.py
@@ -112,6 +112,12 @@ class CUOPT(LpSolver):
                 7: LpStatusNotSolved,  # Primal Feasible
                 8: LpStatusNotSolved,  # Feasible Found
                 9: LpStatusNotSolved,  # Concurrent Limit
+                10: LpStatusNotSolved,  # Work Limit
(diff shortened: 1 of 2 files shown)


ramakrishnap-nv:
May I get review on this ?

ramakrishnap-nv:
@pchtsp May I get your help ?

PR #906, opened 2026-05-05
Update Xpress solver integration for API v9.8 with backward compatibility

brunoficoxpress:
Modernizes the Xpress solver integration to support the current API (v9.8) while remaining backward compatible with older versions.

See also: companion PR for the 3.3.1 branch: #905

### Key Changes
- **XPRESS (CMD):** `available()` now warns if the Xpress Python package is installed but the binary is not found, suggesting `XPRESS_PY` instead
- **XPRESS_PY:** All deprecated lowercase API calls replaced with camelCase equivalents (`optimize`, `lpOptimize`, `writeProb`, `addMessageCallback`, `loadMipSol`, `addMipSol`, `loadLPSol`, `chgObjSense`, `addCols`) with `AttributeError` fallbacks for older versions
- **XPRESS_PY:** Unified MIP/LP status reporting via `solstatus` attribute with `xpress.SolStatus` enum constants and integer fallbacks
- **Docs:** Added XPRESS and XPRESS_PY solver configuration sections to `how_to_configure_solvers.rst`

### Testing
✅ All 73 XPRESS_PyTest tests passed
✅ Tested with Xpress 9.8
✅ Backward compatible with Xpress 9.0+

### DIFF
--- a/pulp/apis/xpress_api.py
+++ b/pulp/apis/xpress_api.py
@@ -29,6 +29,7 @@ from __future__ import annotations
 import math
 import re
 import sys
+import warnings
 from typing import TYPE_CHECKING, Any
 
 from .. import constants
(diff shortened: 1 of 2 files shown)


pulp/apis/xpress_api.py
```diff
@@ -104,7 +105,20 @@ def defaultPath(self):
...
+            return True
+        try:
+            import xpress  # noqa: F401
+
+            warnings.warn(
+                "Xpress optimizer binary not found. "
+                "Consider using XPRESS_PY instead: pip install xpress",
+                UserWarning,
+                stacklevel=2,
+            )
```
pchtsp: in case of ImportError, I don't think the warning will show. Maybe you need to take out the `import xpress` line? So that it doesn't check on whether it's installed or not?

brunoficoxpress: Thank you Franco, good point. Removed the `import xpress` check so the warning always fires when the binary is not found.

Issue #894, opened 2026-01-09
pulp reports Infeasible for Unbounded HiGHS MIPs


robertHowlett:
Details for the issue
--------------------
Pulp incorrectly reports infeasible for Primal infeasible/Unbounded MIPs solved with HiGHs. Unbounded LPs work fine.

#### What did you do?
I ran the script:

```
import pulp

# Problem formulation
m = pulp.LpProblem("myProblem", pulp.LpMinimize)
var_cont = pulp.LpVariable("var_cont", lowBound=None, upBound=None, cat=pulp.LpContinuous)
var_int = pulp.LpVariable("var_int", lowBound=0, upBound=1, cat=pulp.LpInteger)
m += var_cont + var_int

m.solve(
    pulp.HiGHS(
        timeLimit=30,
        gapRel=0.01,
        threads=1,
        msg=True,
        log_to_console=True,
    )
)

print("Status:", pulp.LpStatus[m.status])
assert m.status == pulp.LpStatusUnbounded
```

#### What did you expect to see?
```
<some solver logs>
Status: Unbounded
```

#### What did you see instead?

```
Running HiGHS 1.12.0 (git hash: 755a8e0): Copyright (c) 2025 HiGHS under MIT licence terms
MIP has 0 rows; 2 cols; 0 nonzeros; 1 integer variables (1 binary)
Presolving model
Presolve: Primal infeasible or unbounded

Src: B => Branching; C => Central rounding; F => Feasibility pump; H => Heuristic;
     I => Shifting; J => Feasibility jump; L => Sub-MIP; P => Empty MIP; R => Randomized rounding;
     S => Solve LP; T => Evaluate node; U => Unbounded; X => User solution; Y => HiGHS solution;
     Z => ZI Round; l => Trivial lower; p => Trivial point; u => Trivial upper; z => Trivial zero

        Nodes      |    B&B Tree     |            Objective Bounds              |  Dynamic Constraints |       Work
Src  Proc. InQueue |  Leaves   Expl. | BestBound       BestSol              Gap |   Cuts   InLp Confl. | LpIters     Time

         0       0         0   0.00%   -inf            inf                  inf        0      0      0         0     0.0s

Solving report
  Status            Primal infeasible or unbounded
  Primal bound      inf
  Dual bound        -inf
  Gap               inf
  P-D integral      0
  Solution status   -
  Timing            0.00
  Max sub-MIP depth 0
  Nodes             0
  Repair LPs        0
  LP iterations     0
[...]

PR #910, opened 2026-05-15
Return undefined status for infeasible or unbound

SteefCoder:
Hopefully a fix for #894


pchtsp:
thanks. the changes make sense, you probably need to adapt the unit tests

PR #918, opened 2026-06-02
HiGHS_PY: Support logPath for setting log file

### LINKED ISSUE #917: HiGHS_PY doesn't support logPath
Details for the issue
--------------------

#### What did you do?

Tried to set logPath="my_log.file" with HiGHS_PY

Logs written to "my_log.file"

Logs not written to "my_log.file"

Useful extra information
-------------------------

HiGHS_CMD does correctly set the log file using logPath, it's just HiGHS_PY that doesn't support it. I'll make a PR shortly.

- [X] Linux: ( distro: Debian )
- [ ] Mac OS: ( _version:_ ___ )
- [ ] Other: ___

- [ ] 3.10
- [ ] 3.11
- [ ] 3.12
- [X] 3.13
- [ ] Other: ___

#### I installed PuLP via:

- [X] pypi (python -m pip install pulp)
- [ ] Other: ___ (conda?)

#### Did you also

- [X] Tried out the latest github version: https://github.com/coin-or/pulp
- [X] Searched for an existing similar issue: https://github.com/coin-or/pulp/issues?utf8=%E2%9C%93&q=is%3Aissue%20

WPettersson:
Fixes HiGHS_PY so that the user can set the log file with the pulp-standard logPath parameter.

Closes: https://github.com/coin-or/pulp/issues/917

### DIFF
--- a/pulp/apis/highs.py
+++ b/pulp/apis/highs.py
@@ -309,6 +309,7 @@ class HiGHS(LpSolver):
             gapRel=None,
             threads=None,
             timeLimit=None,
+            logPath=None,
             callbacksToActivate: Optional[list[highspy.cb.HighsCallbackType]] = None,
             **solverParams,
         ):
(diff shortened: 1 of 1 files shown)


pchtsp:
thanks, can you check how we take and store logPath in the other APIs (cbc, gurobi, cplex) so that it matches that pattern?
`logPath` usually is stored inside self.optionsDict by default and is accessed like so:

```
self.optionsDict.get("logPath")
```

WPettersson:
Sorry, I did miss that. I was just following the HiGHS_PY class, but it seems to be the one that does things differently to the other solvers. It's also calling `super()` rather than `LpSolver.__init__()`, and it already had code lines over length 80 so I thought long lines would be okay for now, which is why I've left it as a 100 column line to call `super()`. I can combine it with style changes to shorten line lengths and call `LpSolver.__init__()` directly if you wish.

WPettersson:
Seems like CI also pulled me up on the style, so I'll fix that at least.

PR #940, opened 2026-09-19
Report why the solver stopped separately from what it returned

andig:
Implements the proposal in #382 for the 4.0 line: PuLP now reports *why the solver stopped* and *what it returned* as two separate statuses, and no longer calls a run that hit a limit "Optimal".

### API

* `LpProblem.getSolverStatus()` returns a key of `LpStatus`: why the solver stopped. Three new codes join the existing ones: `LpStatusTimeLimit` (2), `LpStatusMemoryLimit` (3) and `LpStatusNodeLimit` (4), the numbering from the old `change_status` branch.
* `LpProblem.getSolutionStatus()` returns a key of `LpSolution`: whether the solver handed back an optimal solution, a feasible one, none, or a proof that none exists. Same values as the existing `sol_status` attribute.
* Reading `LpProblem.status` warns with `DeprecationWarning` and points at the two methods. Writing it stays silent so solver interfaces and pickles keep working. `solve()` still returns the solver status as before.
* `LpStatusToSolution`, the default used when a solver interface passes no solution status, maps `LpStatusNotSolved` and `LpStatusUndefined` to `LpSolutionNoSolutionFound` instead of `LpSolutionInfeasible` ("No Solution Exists"), which claimed a proof no solver had given.

### Solver interfaces

Every interface that used to report `LpStatusOptimal` for a feasible-but-unproven result now reports the limit it hit, or `LpStatusNotSolved` when the solver does not say, with the solution as `LpSolutionIntegerFeasible`:

| Interface | Change |
| --- | --- |
| `COIN_CMD` | `Stopped on time - objective value …` becomes `LpStatusTimeLimit`; other `Stopped on …` lines become `LpStatusNotSolved` (CBC prints `Stopped on iterations` for node, gap and solution limits alike). Parsing moved into `COIN_CMD.parse_status` so it can be unit tested. |
| `HiGHS_CMD`, `HiGHS` | time and memory limits mapped; iteration, solution, objective bound/target and interrupt no longer read as optimal. A limit hit before the first incumbent (infinite objective) reports no solution. |
| `GUROBI` | `TIME_LIMIT`, `NODE_LIMIT`, `MEM_LIMIT` mapped; `SolCount` decides the solution status. |
| `CPLEX_CMD` | codes 104–113 no longer optimal; 111 (memory limit, feasible) was missing and raised "Unknown status". |
| `CPLEX_PY` | time, node and memory limits mapped for both the feasible and no-solution variants; `MIP_time_limit_infeasible` was reported as `LpStatusInfeasible` although it only means "no incumbent yet". |
| `GLPK_CMD`, `PYGLPK` | `INTEGER NON-OPTIMAL` / `GLP_FEAS` are `LpStatusNotSolved` with a feasible solution; glpsol does not record why it stopped. |
| `SCIP_CMD`, `SCIP_PY` | time, node and memory limits mapped; the "change status to Optimal by convention" override is gone. |
| `XPRESS_PY` | a `FEASIBLE` solution is `LpStatusNotSolved`, refined to time or node limit from `stopstatus`. |
| `COPT_CMD`, `COPT_DLL`, `COPT` | node limit and timeout mapped; `HasMipSol` decides the solution status. The CMD and DLL paths assigned `lp.status` directly and bypassed `assignStatus`. |
| `CUOPT` | time limit mapped; primal feasible / feasible found carry a solution. |
| `CHOCO_CMD`, `CPSAT` | `SATISFIABLE` / `FEASIBLE` are `LpStatusNotSolved` with a feasible solution. |
| `SAS94`, `SASCAS` | `TIME_LIM_*`, `NODE_LIM_*`, `OUTMEM_*` mapped; the `*_SOL` variants carry a solution. |
| `MOSEK` | `prim_feas` carries a solution. |
| `YAPOSIB` | `limitreached` was reported as infeasible. |

`GUROBI_CMD`, `COPT_CMD`, `MIPCL_CMD` and `FSCIP_CMD` read solution files that do not state why the solver stopped, so they are unchanged.

### Docs and examples

README, the module docstring, the constants and `LpProblem` reference pages, the blending case study and the examples use `getSolverStatus()`. HISTORY has an entry under 4.0.0a13.

### Tests

`test_pulp.py` covers the two accessors, the deprecation warning and the default solution status for every solver status. `test_coin_cmd.py` covers `parse_status` against the header lines CBC 2.10 and master write. The existing solver suites pass unchanged: every model they solve to completion is still `LpStatusOptimal`.

### Breaking change

Code that checks `prob.status == LpStatusOptimal` (or `LpStatus[status] == "Optimal"`) after a time-limited solve will now see the limit instead. That is the point of the change, and why it targets 4.0.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


vepain:
Coded with AI but not able to respect the basic of Python function naming. Python conventions are not Java's ones.

andig:
Kind response, thank you.

vepain:
I don't have to be kind with AI. But I respect programmers who try their best and may do some errors.
The proposed code mimics the gurobipy API (they code in a Java style). Some editors do not expect you use Java conventions in Python - that's why we have coding conventions.

andig:
> I don't have to be kind with AI. 

Not at all.

> But I respect programmers who try their best and may do some errors.

It surely doesn't feel this way. Much less so since this coding style is already present:
- getSolverStatus https://github.com/andig/pulp/blob/770bf18284868bd6ec16d2a5fe17ab4de892929a/pulp/core/lp_problem.py#L989
- getSolutionStatus https://github.com/andig/pulp/blob/770bf18284868bd6ec16d2a5fe17ab4de892929a/pulp/core/lp_problem.py#L996

You do realize that it's people triggering the AI and comment in the thread, do you?

pchtsp:
hey. I'm joining late to this discussion. I've actually been playing with a more sophisticated status object being returned after each solve: https://github.com/coin-or/pulp/tree/solve_status

It's very much WIP as haven't even yet looked at what the IA proposed. But given the current discussion, I wanted to flag it and share the fact that I'm thinking on going a different route. What I want is to offer as much information as possible to the user: progress logs, gaps, root node cuts, etc.

andig:
I’d be happy with any solution 👍🏻

PR #942, opened 2026-09-21
LpSolveStats

pchtsp:


andig:
Working for me. 

Claude found one issue that I can only relay verbatim (feel free to ignore): best_bound sign flipped for maximize problems on COIN_CMD

COIN_CMD writes maximize problems as a negated-objective MPS with no OBJSENSE and no -max. CBC's log is therefore in minimize sense. LpSolver.buildStats copies orloge's best_bound and best_solution without flipping them for lp.sense. Minimal repro:

```
p = pulp.LpProblem("m", pulp.LpMaximize)
x = p.add_variable("x", 0, 10, cat="Integer"); y = p.add_variable("y", 0, 10, cat="Integer")
p += 3*x + 2*y; p += x + y <= 7; p += x - y <= 2
st = p.solve(pulp.COIN_CMD(msg=False))
# objective 18.0  best_bound -18.0  gap_abs 36.0  gap_rel 2.0
```

Same model as minimize gives best_bound -18.0, gap_rel 0.0. Fix location: buildStats (or COIN_CMD.actualSolve) must negate log-derived bound and objective when lp.sense == LpMaximize. Other CMD solvers that write LP files with a Maximize section are not affected by this path.

pchtsp:
thank you. Claude is correct about the error detection. orloge is parsing correctly the CBC logs but pulp needs to flip those values when sense == LpMaximize

vepain:
Globally a better idea to return an object after solve instead of changing the LpProblem attributes. Pure frozen dataclasses should be used to be close to a DDD architecture.

Some minor notes:

Try to avoid repeating string literals "Optimal" etc.: use constants instead or enum.StrEnum (as you did for StrInt `LpSolveStatus`)

Method naming conventions in Python is snakecase, e.g. prefer using `LpSolveStats.to_json` instead of `LpSolveStats.toJson`

Do you want to let users adding their own solver Pulp API (plugin system)? If not, prefer using StrEnum, constants or solver services' class attributes as dictionary keys in LOG_DIALECTS  dictionary. This will enforce typing verification.

pchtsp:
thanks for the comments @vepain 

> Try to avoid repeating string literals "Optimal" etc.: use constants instead or enum.StrEnum (as you did for StrInt LpSolveStatus)  

I'm not sure where do you see these. Can you give me an example?

> Method naming conventions in Python is snakecase, e.g. prefer using LpSolveStats.to_json instead of LpSolveStats.toJson  

I agree, I'm usually conflicted between following python convention and this repo's convention (which precedes me and has "toJson" convention for the most part) for new methods. I never know what to follow.

> Do you want to let users adding their own solver Pulp API (plugin system)? If not, prefer using StrEnum, constants or solver services' class attributes as dictionary keys in LOG_DIALECTS dictionary. This will enforce typing verification.  

Ideally, yes: there is even some minor functionality that allows users to plug-in specific solvers (see https://github.com/coin-or/pulp/blob/master/pulp/apis/__init__.py#L103). This doesn't happen very often but someone did ask for this method this year to do exactly that.

vepain:
> I'm not sure where do you see these. Can you give me an example?

Never mind, I get confused while reading. The comment concerned the API but because it is specified only once, it is overkill to create a constant or a StrEnum.

